PR #6, opened 2021-08-27
Extrinsic call

bobo-kovacevic:
Found and issue when submitting extrinsic call with same amount to the same account multiple times. First transaction passes, others not. I need to restart the app if I want to submit second transaction

https://user-images.githubusercontent.com/8452361/131121425-ea43ee63-cc9f-4dc5-9f36-53a5ea22e51f.mp4


sirius651:
LGTM. the above issue could be separate branch to resolve, so now merge with main branch

PR #9, opened 2021-09-09
Feature/custom signature improvement

bobo-kovacevic:
Update extrinsic transfer function to match latest changes on a chain pallet. Supported only on Shibuya at the moment


src/config/api/polkadot/connectApi.ts
```diff
@@ -61,7 +61,7 @@ export async function connectApi(endpoint: string, networkIdx: number) {
   let extensions: InjectedExtension[] = [];
 
   let typeDefinitions = null;
-  if (networkIdx === endpointKey.SHIDEN) {
+  if (networkIdx === endpointKey.SHIDEN || networkIdx === endpointKey.SHIBUYA) {
```
hoonsubin, line 64: instead of having a list of `if` conditions, what about embedding the type definitions on the `ChainProvider` object? https://github.com/PlasmNetwork/astar-apps/pull/9/files#diff-bbb15aefae956e4e9ba6a9527a96e3f2693a2d247e6c2235d13d8f25f3274259R21

for example, we can add a new property to the interface
```ts
interface ChainProvider {
  networkAlias: string;
  displayName: string;
  info?: string;
  // ...
  typeDef: RegistryTypes;
}
```

src/components/balance/modals/ModalTransferAmount.vue
```diff
@@ -249,6 +254,30 @@ export default defineComponent({
       }
     }
 
+    const getPayload = async (method: SubmittableExtrinsic<'promise'>, nonce: u32): Promise<Uint8Array> => {
```
hoonsubin, line 257: having this function directly in the component will make it difficult to reuse client extrinsic payloads. I suggest separating this into a separate function so that it can be used in multiple components.

src/components/balance/modals/ModalTransferAmount.vue
```diff
@@ -12,7 +12,7 @@
         class="tw-inline-block tw-bg-white dark:tw-bg-darkGray-900 tw-rounded-lg tw-px-4 sm:tw-px-8 tw-py-10 tw-shadow-xl tw-transform tw-transition-all tw-mx-2 tw-my-2 tw-align-middle tw-max-w-lg tw-w-full"
       >
         <div>
-          <q-banner v-if="isShidenChain" dense rounded class="bg-orange text-white tw-mb-4 q-pa-xs" style="">
+          <q-banner v-if="!isExtrinsicSupported" dense rounded class="bg-orange text-white tw-mb-4 q-pa-xs" style="">
```
hoonsubin, line 15: the name `isExtrinsicSupported` is too broad. Please use something like `isCustomSigBlocked` and make a note that this is just temporary in a comment.

hoonsubin (changes requested):
Including some changes, please fix the folder name for https://github.com/PlasmNetwork/astar-apps/tree/main/src/config/api/polkadot/registry_types as well. It should be `registry-types` or something like that

hoonsubin (approved):
LGTM

PR #12, opened 2021-09-17
Add repo deploy action

hoonsubin:
- adds project script
- adds PR check action
- adds GH pages deploy action


sirius651:
LGTM. but, when deploying to gh-pages, it seems that we should set custom domain as 'portal.astar.network' manually.

PR #15, opened 2021-09-30
Feature/dapps store

bobo-kovacevic:


.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
bobo-kovacevic, line 1: Commited this by, mistake. I will remove and add to git ignore

service/src/routes/store.ts
```diff
@@ -0,0 +1,55 @@
+import express, { Request, Response, NextFunction } from 'express';
+import { StatusCodes } from 'http-status-codes';
+import * as admin from 'firebase-admin';
+
+const router = express.Router();
+
+admin.initializeApp({
+  credential: admin.credential.applicationDefault(),
+  //databaseURL: 'https://dapp-store-7f9ab.firebaseapp.com'
+  databaseURL: 'https://dapp-store-7f9ab-default-rtdb.europe-west1.firebasedatabase.app/'
```
hoonsubin, line 10: We should store this as an environmental variable

src/components/dapps/modals/InputFile.vue
```diff
@@ -5,6 +5,7 @@
   >
     <div class="tw-space-y-1 tw-text-center">
       <icon-base
+        v-if="!!!$slots.default"
```
hoonsubin, line 8: is `!!!` common to use?

src/hooks/firebase/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  collection,
+  getDocs,
+  setDoc,
+  doc,
+} from 'firebase/firestore/lite';
+import { getStorage, ref, uploadString, getDownloadURL } from 'firebase/storage';
+import { DappItem, NewDappItem } from 'src/store/dapps-store/state';
+
+// firebase init - add your own config here
+const firebaseConfig = {
```
hoonsubin, line 13: the config should be stored as an environmental variable

hoonsubin (changes requested):

src/components/store/Dapp.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <div class="tw-border tw-m-4 tw-w-72 tw-rounded-lg">
+    <div class="tw-flex tw-flex-grow tw-cursor-pointer tw-p-4" @click="emitClickEvent">
+      <Avatar :url="dapp.iconUrl" class="tw-w-14 tw-h-14" />
+      <div class="tw-ml-4">
+        <div class="tw-text-lg tw-font-semibold">{{ dapp.name }}</div>
```
hoonsubin, line 6: Screen Shot 2021-10-03 at 1 36 14 PM
The font color do not change in dark mode

src/components/store/modals/ModalDappDetails.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <Modal>
+    <template v-slot:content>
+      <div class="tw-flex tw-flex-col tw-justify-center tw-items-center">
```
hoonsubin, line 4: The style won't change in dark mode
Screen Shot 2021-10-03 at 1 36 25 PM

hoonsubin (changes requested):

src/components/dapps/modals/InputFile.vue
bobo-kovacevic: Actually, it isn't. I exaggerated with ! :)

src/hooks/firebase/index.ts
bobo-kovacevic: This code will be removed as soon as we switch to the admin service.

service/src/routes/store.ts
bobo-kovacevic: I agree. If I understood correctly to define env. variables we need firebase functions feature enabled which requires paid plan. This leads me to a question, are we going to have Astar firebase account?

hoonsubin: Probably yes. Let’s create one (or use an alternative) when we deploy the dApps Store to Shiden after the Shibuya test. Since for Shibuya, we’ll only need to save contracts from a whitelist.

hoonsubin (approved):
LGTM

Issue #49, opened 2021-10-07
Address display/select cross page


bLd75:
Currently, the address selection/balance view is only available from the balance page.

It would be nice to have this box (including the balance) on every pages, on the top right for example.
Adding the verified identity would also be a useful extra.

![image](https://user-images.githubusercontent.com/18489502/136472923-f5272039-d71b-410b-913d-0aef0c0abe78.png)

hoonsubin:
fixed with https://github.com/PlasmNetwork/astar-apps/pull/68 and https://github.com/PlasmNetwork/astar-apps/pull/63

Issue #48, opened 2021-10-07
Switch from wallet in dApp store


fiexer:
**Motivation**
If you use multiple accounts, you have to go first to your balance before seeing staking amount of that wallet.

**Suggested Solution**
Make it possible to easily switch from wallet in the dApp store page

**Alternatives**
/

**Additional Information**
/

hoonsubin:
duplicate of https://github.com/PlasmNetwork/astar-apps/issues/29

PR #68, opened 2021-10-11
Add account name and address to side nav

### LINKED ISSUE #29: Choose account box available on all views: Balance, Store..
If I have multiple accounts I need to go to Balance view to switch my account before I can stake, it would be more user friendly if I can do it right away from Store view.
image
I propose make this change wallet box available on all views. 

Also if I refresh the browser page from Store view it's navigating me to the same Store view, but my account changed to the first one which is available in my polkadot.js chrome extension, that's why it would be good to see which account is currently selected for staking  on Store view.

sirius651:
Fixed #29 #48 #49


sirius651:
Resolved conflict & set hidden overflow-x. but need look over `overflow-y: auto` is not working properly (scroll is not hidden automatically when no user action)

Issue #74, opened 2021-10-13
Add network status indication


bobo-k2:
**Motivation**
Portal users don't have information about network status and can't know if their transaction will pass or not.

**Suggested Solution**
Add the current block time under the network status bar and mark it red if the block time is more than 10 times the average

sirius651:
what would be average of block time generally?

bobo-kovacevic:
> what would be average of block time generally?

I am also unsure, @hoonsubin  mentioned it in one conversation and I put it to the ticket. Sounds to me like we need to subscribe for new blocks and calculate average time between blocks

hoonsubin:
The average block time can be found here https://github.com/PlasmNetwork/Astar/blob/6573df42d5aa5af8b51e35c1b09fbaa5392f4f97/runtime/shibuya/src/lib.rs#L62
The minimum block time can be queried via `api.const.timestamp.minimumPeriod`
![Screen Shot 2021-11-17 at 9 37 49 AM](https://user-images.githubusercontent.com/40356749/142165808-daca91dc-00a0-428c-be68-fed70a48fdd4.png)

impelcrypto:
Hi @bobo-k2 

Is this issue still status as 'Open'? By the way, this is the code to calculate average block time.

https://github.com/AstarNetwork/astar-apps/blob/main/src/hooks/chain/useAvgBlockTime.ts#L109

bobo-k2:
Closing as fixed

PR #158, opened 2021-11-18
Feat/74 network status

sirius651:
Implement #74 

**Check list**
- [x] add useBlockTime hook
- [x] add block timer component


hoonsubin:
@sirius651 can you revisit this PR at some point? The goal is to add a better network status indicator for the user. For example, right now, when the network is disconnected, the user must look at the browser console log to see what is happening since the UI won't display anything.

sirius651:
> @sirius651 can you revisit this PR at some point? The goal is to add a better network status indicator for the user. For example, right now, when the network is disconnected, the user must look at the browser console log to see what is happening since the UI won't display anything.

sure, I'll check

hoonsubin:
closing this due to portal v2

PR #164, opened 2021-11-21
feat: Added APR calculation

impelcrypto:
**Pull Request Summary**

Added APR calculation

image

**Check list**
- [x] adds new feature
- [x] relies on other tasks
- [ ] documentation changes


hoonsubin:
convert this PR as a draft PR if you're still working on it

src/hooks/useApr.ts
```diff
@@ -0,0 +1,121 @@
...
+          apiRef.rpc.chain.getHeader(),
+        ]);
+
+        const rawTotalSupply = results[0];
+        const totalSupply = reduceBalanceToDenom(rawTotalSupply, decimalRef);
+        console.log('totalSupply:', totalSupply.toString()); // -> 71,001,295 SDN
+
+        const rawBlockRewards = results[1];
+        const blockRewards = Number(defaultAmountWithDecimals(rawBlockRewards, decimalRef));
+        console.log('blockRewards:', blockRewards.toString()); // -> 2.664 SDN
```
Maar-io, line 75: the block reward of 2.664 SDN is divided to maintenance and dAppStaking in 50-50 ratio. Meaning, dAppsBlockReward = 1.332 SDN

src/hooks/useApr.ts
```diff
@@ -0,0 +1,121 @@
+import { useStore } from 'src/store';
+import { computed, ref, watchEffect } from 'vue';
+import { useChainMetadata, useCurrentEra, useTvl } from '.';
+import { defaultAmountWithDecimals, reduceBalanceToDenom } from './helper/plasmUtils';
+import { useApi } from './useApi';
+
+// fixme: ideally get the value from API
+const STAKER_BLOCK_REWARD = 0.1;
+const DAPPS_BLOCK_REWARD = 0.4;
```
Maar-io, line 9: fetch this from chain

src/hooks/useApr.ts
```diff
@@ -0,0 +1,101 @@
...
+import { useChainMetadata, useCurrentEra, useTvl } from '.';
+import { defaultAmountWithDecimals, reduceBalanceToDenom } from './helper/plasmUtils';
+import { useApi } from './useApi';
+
+// Ref: https://github.com/PlasmNetwork/Astar/blob/5b01ef3c2ca608126601c1bd04270ed08ece69c4/runtime/shiden/src/lib.rs#L435
+// Memo: 50% of block rewards goes to dappsStaking, 50% goes to block validator
+// Fixme: ideally get the value from API
+const DAPPS_REWARD_RATE = 0.5;
+
+const TS_FIRST_BLOCK = {
```
hoonsubin, line 12: instead of hard-coding all block timestamps for each blocks, I recommend using the `await api.rpc.chain.getBlockHash(1);` and `await api.query.at(blockHash).timestamp.now();` to get the past block timestamp.

src/hooks/useTvl.ts
```diff
@@ -30,9 +30,10 @@ export function useTvl(api: any) {
       };
 
       const priceUsd = async (): Promise<number> => {
-        if (defaultUnitToken.value === 'SDN') {
+        const chainName = await apiRef.runtimeVersion.specName.toString();
```
hoonsubin, line 33: This fetch is used across multiple scripts. Since the chain information does not change that frequently, I think we should optimize this by fetching the information only once and using that.
The same fetch is called:
- https://github.com/PlasmNetwork/astar-apps/blob/6656893b3422a86139c689be6782f553d1ff137f/src/hooks/useChainInfo.ts#L23
- https://github.com/PlasmNetwork/astar-apps/blob/1101eee8bc182fd2046852613590501a2beec85b/src/hooks/useApr.ts#L60
- https://github.com/PlasmNetwork/astar-apps/blob/6656893b3422a86139c689be6782f553d1ff137f/src/hooks/useMetaExtensions.ts#L79

It would be great if we only have to call the context `useChainInfo`

hoonsubin (changes requested):

src/hooks/useTvl.ts
impelcrypto: It seems like `useChainInfo` is not working properly (returns `undefined`). I'll take a look and try to fix it.

src/hooks/useApr.ts
impelcrypto: yes, I was looking for something like this for a very long time. I think `await api.query.at(blockHash).timestamp.now()` is not callable, instead it should be queried like [this](https://github.com/polkadot-js/api/issues/3970#issuecomment-922492767). However, it doesn't work for our (Shiden) API somehow...

hash: 0x985b77fdbc66f3ea8316b7107b86e8542c70dd045eed26f123795903d8e9cba3
 
image

image

image

hoonsubin: This seems like a storage MetaData V4 error that we were seeing before.
Screen Shot 2021-11-25 at 5 58 53 PM
I thought this was solved in https://github.com/PlasmNetwork/astar-apps/pull/167, but maybe not?
@bobo-kovacevic do you know the cause of this?

impelcrypto: @hoonsubin 
How about improving the timestamp part once MetaData issue has been fixed? Then perhaps we can merge this APR feature first?

hoonsubin: yeah, that sounds good

hoonsubin (approved):
LGTM

PR #183, opened 2021-12-05
feat: added 'staked only' filter and search input in Store section

impelcrypto:
**Pull Request Summary**

* Added 'staked only' filter and search input in Store section

Issue: 

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
image

### DIFF
--- a/src/components/store/DappsFilter.vue
+++ b/src/components/store/DappsFilter.vue
@@ -1,0 +1,70 @@
+<template>
+  <div
+    class="
+      tw-flex tw-justify-center
+      md:tw-justify-between
+      tw-items-center tw-flex-wrap tw-gap-x-2
+      sm:tw-gap-x-3
+      md:tw-gap-x-7
+      tw-gap-y-4
+      md:tw-gap-y-0
+    "
+  >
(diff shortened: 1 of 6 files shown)


src/components/store/DappsFilter.vue
```diff
@@ -0,0 +1,70 @@
...
+      tw-items-center tw-flex-wrap tw-gap-x-2
+      sm:tw-gap-x-3
+      md:tw-gap-x-7
+      tw-gap-y-4
+      md:tw-gap-y-0
+    "
+  >
+    <div class="tw-flex tw-flex-col tw-justify-center tw-items-center tw-gap-y-1">
+      <div class="tw-font-semibold tw-text-blue-900 dark:tw-text-white">Staked only</div>
+      <Toggle v-model:value="isToggleOn" :is-small="true" @toggleAction="toggleAction" />
```
hoonsubin, line 15: Adding a toggle button will make it very difficult to make the UI scale since we will have more searches and filters. Instead, what about creating a `Manage` tab or `My Stakes` tab next to the `Discover` tab that only displays all the staked dapps?

hoonsubin (changes requested):

PR #194, opened 2021-12-19
fix: Update wallet page UI

impelcrypto:
**Pull Request Summary**

Update wallet page UI

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Changes**

image

image

image

- [ ] (ex: add user list)


hoonsubin (approved):
LGTM

hoonsubin (changes requested):
Can you add edges cases for users who do not have any accounts on their wallet or do not have a wallet at all? A message popup that contains how to add accounts and where to download the wallet

src/components/balance/Accounts.vue
```diff
@@ -0,0 +1,105 @@
+<template>
+  <div class="tw-flex tw-flex-col md:tw-flex-row tw-gap-y-8 tw-gap-x-8 tw-flex-wrap">
+    <template v-if="isMetamask">
```
hoonsubin, line 3: I don't think this method of displaying is scalable. For example, what if we want to add foreign currencies like XCMP and bridge tokens?
I suggest rendering this from a array of accounts.

hoonsubin:
This PR will be closed due to changing our design direction.

PR #201, opened 2021-12-30
feat: connect wallet

impelcrypto:
**Pull Request Summary**

Changed wallet login method

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Changes**
* The wallet page contains a 'Connect Wallet' button
  * Display 'Install wallet extension' modal if the user clicks on a wallet that is not installed
  * Automatically add/change the network on the MetaMask
  * Default address schema as H160 whenever connected to MetaMask
* Disabled automatically login
* User can connect from MetaMask or Polkadot.js
* Fixed 'update metadata' button

**To-dos**
* Add link to 'Learn how to Connect' button 

image

image

image

- [ ] (ex: add user list)


hoonsubin (changes requested):
bandicam 2021-12-30 15-38-06-896
Please remove the address switch scheme when connecting to MetaMask. We don't need this anymore since all behaviors are on by default.

hoonsubin:
bandicam 2021-12-30 16-32-00-454
The transaction doesn't seem to work when connected to MetaMask?

impelcrypto:
I haven't changed any logic in H160 view. 
Do you want to transfer from H160(MetaMask) to EVM mapped SS58 account?

hoonsubin:
Yes, that should be the correct behavior. The user should be able to send to H160 and SS58.

impelcrypto:
ok, fixed.

hoonsubin (approved):
LGTM
Let's wait for the growth team to make an announcement and we can merge this once we get the greenlight.

SotaWatanabe:
Screenshot 2022-01-03 at 9 31 53 PM

@impelcrypto this is a very small thing but could you fix the text center of the box?

Maar-io:
After I pick Polkadot.js as wallet I get following message. I do already have installer P extension. I would expect to go straight to the page without seeing this modal.
Screenshot 2022-01-03 at 14 51 39

Maar-io:
on top of the issue I mentioned above. I can't get rid of this modal. it stays open forever and I can't access the page
@impelcrypto

impelcrypto:
@SotaWatanabe 
ok, fixed.
image

src/components/balance/Wallet.vue
```diff
@@ -18,16 +18,15 @@
         </div>
       </div>
 
-      <button type="button" class="icon tw-ml-auto tw-tooltip" @click="openModal">
+      <button type="button" class="icon tw-ml-auto tw-tooltip" @click="disconnectAccount">
```
hoonsubin, line 21: Screen Shot 2022-01-04 at 11 17 43 AM

Currently, clicking the logout button will disconnect the account and show the login button.

Screen Shot 2022-01-04 at 11 18 37 AM
How about we display the wallet connection modal after disconnecting the user, so we reduce the number of clicks when the user wants to switch the account?

In other words, instead of click logout -> click login -> click wallet to connect, we just make it click logout -> (disconnect the account) -> automatically show the wallet connect modal

hoonsubin (changes requested):
Adding to my other review:

Screen Shot 2022-01-04 at 11 20 48 AM

Screen Shot 2022-01-04 at 11 16 13 AM

Right now, connecting to the portal with MetaMask will hide the dApps Staking tab since users cannot use it with MetaMask (until the precompiled EVM contract is finished).
But dApps Staking is the portal for Astar ecosystem, so it should be visible at all times (even if the user did not log in or from a phone).

So what about we show a message in the dApps Staking page when the user can interact with it or not (i.e., when logged out or logged in with MM) and prevent the users from signing it with MetaMask, but still allow them to view it.

src/components/balance/Wallet.vue
impelcrypto: Ok, fixed.
https://gyazo.com/7d9a4571b5112a09e2f09871c586201a

impelcrypto: I've added toast and disabled buttons for MetaMask account.
image

impelcrypto: And I have updated icons.

image

hoonsubin (approved):
LGTM

PR #207, opened 2022-01-13
fix: change wallet extension injection to support different substrate wallets 

impelcrypto:
**Pull Request Summary**

* Created [wallets.ts](https://github.com/AstarNetwork/astar-apps/blob/fix/change-injection-method/src/config/wallets.ts) file for supporting wallet configuration 
* Created  helper functions in [wallet.ts](https://github.com/AstarNetwork/astar-apps/blob/fix/change-injection-method/src/hooks/helper/wallet.ts) that exports the selected extension's injector
* Support Clover wallet extension
* Removed `allAccounts` and `allAccountNames`. Added global state called `substrateAccount`(object) that contains both `account address` and `account name`. So the app is no longer needs an index to changing account in the each modals
* Hided 'Contract' tab

**Check list**
- [x] contains breaking changes
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

image


hoonsubin (approved):
LGTM

PR #232, opened 2022-01-21
feat: support dApp staking from Ethereum wallet

impelcrypto:
**Pull Request Summary**

* Support dApp staking from Ethereum wallet (Ledger via MetaMask).
* Refactoring existing custom signature function (transfer and EVM withdraw) 
 
**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)

### DIFF
--- a/src/components/balance/modals/ModalTransferAmount.vue
+++ b/src/components/balance/modals/ModalTransferAmount.vue
@@ -124,7 +124,7 @@ import BN from 'bn.js';
 import FormatBalance from 'components/balance/FormatBalance.vue';
 import InputAmount from 'components/common/InputAmount.vue';
 import { getProviderIndex, providerEndpoints } from 'src/config/chainEndpoints';
-import { useApi, useChainMetadata } from 'src/hooks';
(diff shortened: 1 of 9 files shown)


hoonsubin (approved):
LGTM

I confirm that the custom signature dApps Staking feature works.

However, the code doesn't look that maintainable in my opinion. For example, we have separate functions like https://github.com/AstarNetwork/astar-apps/pull/232/files#diff-1a14dec631df517919caa5178c4be66c3259dd15e72d996191fc25e8e00860d1R191 which is the same call as the normal staking dispatch, but specifically for custom signature calls, even though custom signature calls are just a wrapper for normal calls. This will be a problem when we want to add another extrinsic call with custom signature from a different page since we need to do the same thing for each calls.

But this feature has a high priority, so I will merge this now, but I think we will want to refactor this later.

PR #241, opened 2022-01-27
Enhancement/refactors

sirius651:
**Pull Request Summary**

for this tasks: 
https://www.notion.so/staketechnologies/Refactor-hooks-vuex-store-103dc78054824358987dad1872d2a5f1

- Get rid off PolkadotProvider, ApiLoader
- Use boot file for connecting polkadot api : https://quasar.dev/quasar-cli/boot-files
- Remove unused store actions
- use `watchPostEffect`

Generally, my main idea was that since `ApiLoader` is complex and redundant and the `provider-inject` method is not accessible globally, bootFile is used to define and approach it at the very beginning. Therefore, by referring to `import {$api} from 'boot/api` instead of `useApi`, it is not necessary to pass `apiPromise` to the hook anymore. (like `useTvl($api)`). Currently, refactoring to remove api params from hooks has not been conducted. (maybe you can take this out). it leads to remove PolkadotProvider, ApiLoader.
Regarding `useStore`, I left this is because quasar initially sets up a basic store.


hoonsubin (approved):
LGTM

hoonsubin:
> @sirius651 @hoonsubin Seems like the fade in/out animation is a bit weird on my end. https://gyazo.com/200fe374928fe205352ac4c525898490

shoot, you're right. I'm sorry I only did a code check for this PR and not live testing. This is my bad. Can we fix this issue?

PR #243, opened 2022-02-01
feat: integrate cBridge 

impelcrypto:
**Pull Request Summary**

* Added cBridge widget in /bridge/evm
Staging URL: https://deploy-preview-pr-243.herokuapp.com/#/balance/wallet

Todo: Ask cBridge team to test the functionality

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Add**

image

Pending animation
https://gyazo.com/4a3805c7875fc3bb2816c3d72cd02346

image

Screenshot 2022-02-04 at 1 28 48 PM

image

image

image

image


hoonsubin (changes requested):
There are a couple of UX-related issues.

1. If the bridge is only available on Astar, we should show the menu only when the user is connected to Astar. When I switch the `From` network to Shiden while being connected to Astar network in the portal, my MetaMask will request to switch networks while the portal is still on the same network. This feels very disjointed.
It would make more sense if the `From` network is the current network, but when the user selects a different network as `From`, the `To` network should be the current network. In other words, we don't want users to swap from Astar to Ethereum while being connected to Shiden.

Screen Shot 2022-02-04 at 9 15 58 AM

2. If the bridge is only usable with MetaMask, we should show the menu only when the user is connected with MetaMask.

Screen Shot 2022-02-04 at 9 16 22 AM

3. As a new user, it is confusing to understand how the rate works and what are bridge providers. For example, where can I get or buy AAVE coins, and can I really get 5000 ETH for 5000 ASTR tokens? I know this is just sending ERC20 tokens from one chain to another, but the first thing I see is the native tokens, not ERC20.
Maybe adding the token symbol at the end of the estimated value might help.

Screen Shot 2022-02-04 at 9 21 26 AM

hoonsubin (changes requested):
https://user-images.githubusercontent.com/40356749/152779107-4d251a0b-6298-4198-aa1a-b12eca49cd6b.mov

Every time I select a different network, the page asks me to switch my network from MetaMask

impelcrypto:
@hoonsubin Yes, I thought MetaMask should always ask user to change the `networkId` to match with `source chain`

hoonsubin:
> @hoonsubin Yes, I thought MetaMask should always ask user to change the `networkId` to match with `source chain`

Yes, but I'm already connected to the same network

impelcrypto:
You have selected `Astar` as the source chain at 00:19 but the network in your MetaMask is `Ethereum Mainet` at 00:22 

image

image

sirius651 (approved):
LGTM

bobo-k2 (approved):
LGTM. Great work.

hoonsubin (changes requested):
Found an issue:

https://user-images.githubusercontent.com/40356749/152972590-cc91eeb7-f1ce-4d14-a362-2286418fc1fa.mov

Connecting to Shiden and clicking the network switch button on the bridge from Shiden to Astar disconnects the wallet from the page

impelcrypto:
>Connecting to Shiden and clicking the network switch button on the bridge from Shiden to Astar disconnects the wallet from the page

@hoonsubin I just saw your comment. The portal has to `reload` the page whenever the network has been changed. I think I can't modify this behavior because the initial state for the destination chain is `Astar` and source chain is `Ethereum` whenever the portal selects `Astar network`.
 
https://github.com/AstarNetwork/astar-apps/blob/main/src/components/balance/modals/ModalNetwork.vue#L132
https://github.com/AstarNetwork/astar-apps/blob/8cc982f7f4f1242d3994e7975ec85a3362e11fd0/src/hooks/c-bridge/useCbridge.ts#L187

hoonsubin:
> @hoonsubin Perhaps do not change the portal network whenever clicking the reverse button regardless of whether the 'from' or 'to' network is Astar or Shiden? In that case, user can bridge from Astar EVM while the portal has selected Shiden network.

Will they be able to use the bridge in that case? Another problem I found was when the user selects from Shiden to Astar and clicks on the reverse button, the option will reset to Ethereum to Astar so the portal connection can change to Astar.

impelcrypto:
>Will they be able to use the bridge in that case?

Yes, no problem. MetaMask will ask user to change the network (by popup)

> when the user selects from Shiden to Astar and clicks on the reverse button, the option will reset to Ethereum to Astar so the portal connection can change to Astar.

Yes, actually I covered [it](https://github.com/AstarNetwork/astar-apps/pull/243#issuecomment-1032665513) here. That's because the page has reloaded, but I can fix this in [this](https://github.com/AstarNetwork/astar-apps/pull/243#issuecomment-1032682247) solution.

impelcrypto:
@hoonsubin Fixed.

hoonsubin (approved):
LGTM

We will merge this once the announcement has been made

src/config/web3/utils/transactions.ts
```diff
@@ -4,7 +4,7 @@ export const getBalance = async (web3: Web3, address: string) => {
   return await web3.eth.getBalance(address);
 };
 
-export const sendTransaction = async (
+export const sendNativeTokenTransaction = async (
```
impelcrypto, line 7: @sirius651 
FYI: I'd like to modify the function name here.

hoonsubin:
@impelcrypto please resolve the conflicts

PR #244, opened 2022-02-02
fix: Updated TVL logic 

impelcrypto:
**Pull Request Summary**

* Updated TVL logic due to backend migration

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Fixes**

=Before=
image

=After=
image

image


src/hooks/useApr.ts
```diff
@@ -86,7 +87,7 @@ export const useApr = () => {
         const avgBlocksPerDay = avrBlockPerMins * 60 * 24;
         const dailyEraRate = avgBlocksPerDay / blocksPerEraRef;
         const annualRewards = eraRewards * dailyEraRate * 365.25;
-        const totalStaked = Number(reduceBalanceToDenom(tvlTokenRef, decimalRef));
+        const totalStaked = Number(ethers.utils.formatUnits(tvlTokenRef.toString(), decimalRef));
```
hoonsubin, line 90: can you tell me what is the difference between using `ethers.utils.formatUnits` and `defaultAmountWithDecimals`? The formatting seems to be the same, I would like to know why we should use `ethers` package.

impelcrypto: `const tvl = result.unwrap().staked.valueOf();`

Seems like [tvl](https://github.com/AstarNetwork/astar-apps/pull/244/files#diff-f0a5212a1722e187d121ec918aa3179f7fc51c7adb806cc63e9bc3018431f9e2L31) was a BN type before migration but it has been changed. `reduceBalanceToDenom` only takes BN as the first argument so that is why the TVL hasn't been displayed on UI. ([Error message](https://gyazo.com/ef4e5dcda133183a07e74524b8bf1e93))
As long as the tvl can be cast to string type after migration, we can use `ethers.utils.formatUnits` to get the converted value. I'd like to use `ethers` package because of:

1: 
I've spent a lot of time finding the best way to parse/cast the amount coming from ERC20 token. [ethers](https://docs.ethers.io/v5/api/utils/display-logic/) contains `ethers.utils.formatUnits` and `ethers.utils.parseUnits` which work really well. I believe this is the simplest and most accurate (rounding up or down due to decimals) way. 
(I think even though we can avoid using our own functions except for some special cases like round down)

2: 
I have to use this package in cBridge integration 

=`tvl` Before migration (Astar)=
image

=`tvl` After migration (Shiden)=
image

hoonsubin (approved):
LGTM

PR #248, opened 2022-02-08
Enhancement/web3 optimal

sirius651:
**Pull Request Summary**

make the code scalable while using Web3.js and Polkadot-js/api separately
related with task: Implement dApps Staking via Ethereum Wallet

**Check list**
- [ x ] contains breaking changes
- [ x ] adds new feature
- [ x ] modifies existing feature (bug fix or improvements)
- [ x ] relies on other tasks

- refactor transfer page by creating with `useTransfer`
- refactor `useBalance` hook
- separate web3 related function from main hooks
- move `web3` package into in 'config' directory
- add web3 contract module for providing the task as `Implement dApps Staking via Ethereum Wallet`. `src/config/web3/contracts`
- add `useStaking` hook


src/config/web3/utils/convert.ts
```diff
@@ -0,0 +1,51 @@
...
+import { ASTAR_SS58_FORMAT } from 'src/hooks/helper/plasmUtils';
+
+// Memo: The EVM address won't be same as the address shown in MetaMask imported from the same private key of the SS58
+// Ref: https://github.com/polkadot-js/common/issues/931
+export const toEvmAddress = (ss58Address: string) => {
+  return u8aToHex(addressToEvm(ss58Address));
+};
+
+export const checkSumEvmAddress = (evmAddress: string): string => {
+  const web3 = new Web3();
```
hoonsubin, line 14: The Web3 object is being instantiated multiple times just to call the utils function. I recommend using a different library for doing this.

hoonsubin (changes requested):

src/components/balance/modals/ModalTransferAmount.vue
```diff
@@ -196,131 +186,26 @@ export default defineComponent({
...
-      }
-    };
+    const { callTransfer } = useTransfer(selectUnit, decimal, closeModal);
 
-    const transfer = async (transferAmt: number, fromAddress: string, toAddress: string) => {
-      // console.log('transfer', transferAmt);
-      // console.log('fromAccount', fromAddress);
-      // console.log('toAccount', toAddress);
+    const transfer = () => {
+      // console.log('transfer', transferAmt.value);
```
bobo-k2, line 195: Maybe to remove commented code

bobo-k2 (approved):

src/config/web3/utils/convert.ts
roy1210: @sirius651 @hoonsubin 
How about use [ethers.utils. getAddress](https://docs.ethers.io/v5/api/utils/address/#utils-getAddress) instead? `ethers` doesn't required for creating the new instance here.

impelcrypto: @sirius651 @hoonsubin
How about use [ethers.utils.getAddress](https://docs.ethers.io/v5/api/utils/address/#utils-getAddress) instead? `ethers` doesn't required for creating the new instance here.

```suggestion
export const checkSumEvmAddress = (evmAddress: string): string => {
  return ethers.utils. getAddress(evmAddress)
}
```

sirius651: thanks. fixed it based on your suggestion. but I'm curious isn't it better to only use `either` or `web3` instead of both?

impelcrypto: That's a good point. I like to use some utils that come with `ethers.js`, but more familiar to use `web3.js` whenever working with `wallet provider`.

impelcrypto (approved):

src/config/web3/utils/convert.ts
sirius651: okay, that would be another task atm.

src/config/web3/utils/convert.ts
```diff
@@ -0,0 +1,53 @@
...
+  // Memo: returns `false` if evmAddress was converted from SS58
+  try {
+    ethers.utils.getAddress(evmAddress);
+  } catch (e) {
+    return false;
+  }
+
+  try {
+    // Memo: check if the given evmAddress is convertible
+    const ss58Address = toSS58Address(evmAddress);
```
hoonsubin, line 29: @sirius651 is there a reason why this is inside a `try` block? I don't think any of the functions here throws anything

src/config/web3/utils/convert.ts
```diff
@@ -0,0 +1,53 @@
...
+    return false;
+  }
+};
+
+export const toSS58Address = (h160Address: string) => {
+  const address = checkSumEvmAddress(h160Address);
+  return evmToAddress(address, ASTAR_SS58_FORMAT);
+};
+
+export const buildEvmAddress = (toAddress: string) => {
```
hoonsubin, line 43: There are scripts that use `toEvmAddress` directly instead of `buildEvmAddress`. Please unify this.
- https://github.com/AstarNetwork/astar-apps/blob/8f5eada80465898395094c5f779dc422139fde54/src/components/balance/Address.vue#L188
- https://github.com/AstarNetwork/astar-apps/blob/8f5eada80465898395094c5f779dc422139fde54/src/components/balance/modals/ModalWithdrawalEvmDeposit.vue#L181
- https://github.com/AstarNetwork/astar-apps/blob/8f5eada80465898395094c5f779dc422139fde54/src/components/balance/modals/ModalWithdrawalEvmDeposit.vue#L203
- https://github.com/AstarNetwork/astar-apps/blob/8f5eada80465898395094c5f779dc422139fde54/src/hooks/useEvmDeposit.ts#L27
- https://github.com/AstarNetwork/astar-apps/blob/3f1fc5f22b16ec29ff455f1a23422d1b9eac2c77/src/hooks/useTransfer.ts#L73

src/hooks/useBalance.ts
```diff
@@ -29,15 +30,14 @@ function useCall(apiRef: any, addressRef: Ref<string>) {
   const unsub: Ref<VoidFn | undefined> = ref();
 
   const updateAccountH160 = async (address: string) => {
-    if (!address) return;
     try {
       const web3 = await createWeb3Instance(currentNetworkIdx.value as TNetworkId);
```
hoonsubin, line 34: according to this, do I have to create the Web3 instance every time the page updates the h160 account?

hoonsubin (changes requested):

src/config/web3/utils/convert.ts
sirius651: I just copied from other function. let me remove it

src/hooks/useBalance.ts
sirius651: good catch. I refactored as creating web3instance at boot time, it can be accessed by globally.

hoonsubin (approved):
LGTM

PR #250, opened 2022-02-10
Dapp video url fix

bobo-k2:
**Pull Request Summary**
Sometimes it happens that video is not displayed in dapp details dialog. I believe the main reason for that is we are accessing the portal over https and then trying to access video over http. 

image

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Fixes**
- dapp info video url

**To-dos**
> update firebase videos url to use ssl


bobo-k2:
Just updated dapp video urls for Astar and Shiden to use https

PR #252, opened 2022-02-11
fix: added 'Transfer from MetaMask' message

impelcrypto:
**Pull Request Summary**

* added 'Transfer from MetaMask' message
* refactoring SocialMediaLinks.vue  (import urls rather than hardcoding.)

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Changes**
- (ex: Change document B)

Before:
image

After:
image

### DIFF
--- a/src/components/balance/Addresses.vue
+++ b/src/components/balance/Addresses.vue
@@ -8,28 +8,44 @@
       tw-py-4 tw-gap-y-2
     "
   >
-    <Address :format="AddressFormat.SS58" />
-    <div class="tw-px-32">
-      <div class="tw-border-b tw-border-gray-100 dark:tw-border-darkGray-600" />
+    <Address />
+    <div
+      class="
+        tw-border tw-border-gray-300
+        dark:tw-border-white
+        tw-rounded-md tw-p-2 tw-m-3 tw-mb-2 tw-text-blue-900
+        dark:tw-text-darkGray-100
+      "
+    >
+      <p class="tw-font-bold">{{ $t('balance.transferFromMetamask', { token: tokenSymbol }) }}</p>
+      <p>{{ $t('balance.howToSendFromMetamask', { token: tokenSymbol }) }}</p>
+      <a
+        class="tw-underline"
+        target="_blank"
+        rel="noopener noreferrer"
+        :href="docsUrl.evmDeposit"
+        >{{ $t('balance.readTheTutorial') }}</a
+      >
     </div>
-    <Address :format="AddressFormat.H160" />
   </div>
 </template>
 <script lang="ts">
-import { defineComponent } from 'vue';
+import { useStore } from 'src/store';
+import { computed, defineComponent } from 'vue';
 import Address from './Address.vue';
-
-export enum AddressFormat {
-  SS58 = 'SS58',
-  H160 = 'H160',
-}
+import { docsUrl } from 'src/links';
 
 export default defineComponent({
   components: {
     Address,
   },
   setup() {
-    return { AddressFormat };
+    const store = useStore();
+    const tokenSymbol = computed(() => {
+      const chainInfo = store.getters['general/chainInfo'];
+      return chainInfo && chainInfo.tokenSymbol;
+    });
+    return { tokenSymbol, docsUrl };
   },
 });
 </script>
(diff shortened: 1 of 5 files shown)


impelcrypto:
@sirius651 @hoonsubin 
could you help review this PR, please? thank you!

src/components/balance/Addresses.vue
```diff
@@ -8,28 +8,44 @@
...
+    <div
+      class="
+        tw-border tw-border-gray-300
+        dark:tw-border-white
+        tw-rounded-md tw-p-2 tw-m-3 tw-mb-2 tw-text-blue-900
+        dark:tw-text-darkGray-100
+      "
+    >
+      <p class="tw-font-bold">{{ $t('balance.transferFromMetamask', { token: tokenSymbol }) }}</p>
+      <p>{{ $t('balance.howToSendFromMetamask', { token: tokenSymbol }) }}</p>
```
hoonsubin, line 21: I'm a bit confused by this message. Why is there a tutorial for sending tokens from EVM to the native wallet but not one for the native wallet to EVM? Is it because of the EVM deposit?
Maybe I'm missing something, but most users will have native tokens first before using EVM, so showing this under the native address does not make sense to me, especially for users who are not going to use this feature.

I agree with removing the H160 address if the user is logged in with a polkadot wallet, but replacing it with a message doesn't seem good to me.

Also, I think we can change the behavior of the EVM transfer modal to show the warning (https://github.com/AstarNetwork/astar-apps/blob/a8ba0e1aa6d8a784db75cf60587104a1064f39b8/src/components/balance/modals/ModalTransferAmount.vue#L42) and the checkbox (https://github.com/AstarNetwork/astar-apps/blob/a8ba0e1aa6d8a784db75cf60587104a1064f39b8/src/components/balance/modals/ModalTransferAmount.vue#L128) if the user pasted an ss58 address.
 
Screen Shot 2022-02-14 at 2 04 19 PM

If the user pasted an h160 address, no messages should be required.

hoonsubin (changes requested):

src/components/balance/Addresses.vue
Kahonnohak: It's the facts that :
- People use Metamask when they think of EVM wallet but not Portal so we would like to educate people to use Portal and that's easier.

- People don't really understand that you cannot send tokens other than ASTR/SDN from Metamask to Portal H160 (when it is done they lose funds)

- It is good that even though transfer will get an error when trying to send tokens (ASTR/SDN) from Metamask to SS58 but they won't lose their funds (meaning it is not critical)

- Bridge user exists that they bring funds to Astar and swap to get native tokens

Based on all I thought it is safer to have those messages for this stage. First-time user onboarding UI should be added for Portal V2 and will think about how. How do we let people know that they should stay on Portal and change the wallet to Metamask when they want to transfer from Polkadot.js. That was all about what I was thinking.

Am I worrying too much?

hoonsubin: So do you suggest that we add a prompt for first visiting users to read the manual like other portals? Or go with the message below the ss58. I just don't like the asymmetric design where we show different UI based on which wallet the user is connected to. It complicates things.

Since we have a redesign planned anyway, I do not have a strong opinion about this, so I will go with your final thoughts

Kahonnohak: We will have tutorials and could have FAQ. For V2, what I'm thinking so far (not ready for review though) is - when you press the Account button on the header this right bit comes out and you can switch the wallets and accounts. And I'm thinking to put first time message on this area, something like 
- Polkadot wallet to send tokens to CEX
- Connect Metamask wallet to send Native tokens to Polkadot address
- Using our Portal for EVM transfer to avoid any mistakes that might happen (Our Portal only enable transactions when it is possible) 

(Messages are just examples)

No, actually I don't have a strong will for having this message on the portal V1 either. It is better not to have H160 and will avoid accidents. Let's just remove H160 for now.

I do have bunch of questions for transferring and bridging let me share my questions when you are back from  Denver, please! 

account

hoonsubin: We cannot remove H160 because of the upcoming feature which is enabling dApps Staking for H160 accounts. If you believe that adding the message below the ss58 account can prevent issues, I think we can leave it at that

hoonsubin (approved):
We will leave the message for now, but this will change when we redesign the portal

sirius651 (approved):

impelcrypto:
@hoonsubin 
I've removed warning messages that popup whenever transferring from H160 to H160.

PR #258, opened 2022-02-17
feat: individual claim

impelcrypto:
**Pull Request Summary**

Applied upgraded pallet:
* Created `isEnableIndividualClaim -> bool` variable for making [if statement]( query the total pending rewards and number of rewardable dApps ) that invoke the functions and queries to the upgraded pallet
* Updated TVL calculation
* Updated getStakeInfo function
* Added Claim All UI 
* Migrated claim functions (claimStaker, claimDapp)
* Added a toast for displaying how much the staker has claimed ([demo](https://github.com/AstarNetwork/astar-apps/pull/258#issuecomment-1062690949))

image

**Check list**
- [x] contains breaking changes
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


src/config/chainEndpoints.ts
```diff
@@ -127,3 +128,24 @@ export const getProviderIndex = (chain: ASTAR_CHAIN) => {
       return endpointKey.ASTAR;
   }
 };
+
+// Memo: remove this function whenever individual claim  is applied in all the networks
+export const checkIsEnableIndividualClaim = () => {
```
bobo-k2, line 133: I think it would be better to check a node version or read some specific feature of individual claim and base on that calculate checkIsEnableIndividualClaim, rather than have logic which needs to be maintained as individual claim feature is pushed on different networks

impelcrypto: Sounds good. Feel free to modify this function 👍

impelcrypto:
@bobo-k2 
Seems like we still have a problem with pending rewards calculations. Kindly refer to the attached video.
https://www.loom.com/share/4c213fb32965475cbdfa29245b19f226

1. Portal displays `Assertion failed` error on the first dApp 
2. The pending rewards as 9K are wrong. The actual rewards are 11 ASTL.

impelcrypto:
> Portal displays `Assertion failed` error on the first dApp

Screenshot 2022-02-21 at 10 54 30 AM

Screenshot 2022-02-21 at 10 54 11 AM

impelcrypto:
@bobo-k2 
May I know if it is ok to omit `nonce: -1`?

https://github.com/AstarNetwork/astar-apps/blob/73d0eac99c14beec5c7a1a2f84eb8f4d6e5d233d/src/store/dapp-staking/actions.ts#L551 

https://github.com/AstarNetwork/astar-apps/blob/73d0eac99c14beec5c7a1a2f84eb8f4d6e5d233d/src/hooks/useIndividualClaim.ts#L60

impelcrypto:
> [Calculation fix](https://github.com/AstarNetwork/astar-apps/pull/258/commits/4479ac6a9b86cba5cc0b7ad1a3389adbdca33c93)
[4479ac6](https://github.com/AstarNetwork/astar-apps/pull/258/commits/4479ac6a9b86cba5cc0b7ad1a3389adbdca33c93)

@bobo-k2 
Seems like calculating `pendingRewards` doesn't work correctly on my end. How about you?

image

bobo-k2:
I am not getting errors. Can you write steps to reproduce?

impelcrypto:
@bobo-k2 
https://www.loom.com/share/75fe145fc3324ea1b67bdb89d5820d78

>  am not getting errors. Can you write steps to reproduce?

1. perhaps try staking on 2 dApps. You can find this error at 4:11 in this video.

2. I think this is calculation error. The original balance was 605 ASTL and pending rewards was 14.4ASTL (4:23), but the balance after claiming was 610ASTR (4:52).

impelcrypto:
@bobo-k2 
This is another error after pulling the latest version of [this](https://github.com/AstarNetwork/Astar/tree/feature/individual-claim-local) branch.

1. pull the latest version of [this](https://github.com/AstarNetwork/Astar/tree/feature/individual-claim-local) branch (or remove and clone again).
2. run the local node and stake on dapp
3. click `claim`
 
image

src/hooks/helper/claim.ts
```diff
@@ -0,0 +1,287 @@
+import { BatchTxs } from './index';
+import { getAddressEnum } from './../../store/dapp-staking/calculation';
+import { Struct, Option } from '@polkadot/types';
+import { ApiPromise } from '@polkadot/api';
+
+interface ContractEraStake extends Struct {
```
Dinonard, line 6: This has been modified in the latest version:

https://www.notion.so/staketechnologies/dApps-Staking-Individual-Claim-67654302ddb843bc89f505358205fd88#322b0a67fad3466db6ee5e22db5c92ec

src/hooks/helper/claim.ts
```diff
@@ -0,0 +1,287 @@
...
+  const data = await api.query.dappsStaking.stakersInfo<StakersInfo>(
+    senderAddress,
+    getAddressEnum(dappAddress)
+  );
+  if (data && !data.isEmpty) {
+    const stakes = data.stakes;
+    const firstEraToClaimForStaker = Number(stakes[0].toHuman().era);
+    const numberOfUnclaimedEra = currentEra - firstEraToClaimForStaker;
+
+    return { firstEraToClaimForStaker, numberOfUnclaimedEra };
```
Dinonard, line 37: Unfortunately, it's a bit more complicated than this :)

You have to account for possible _holes_ between eras.
Please check [this](https://github.com/AstarNetwork/astar-frame/pull/6/files#diff-74f8acb4bf207fcaa637a96dcba0e2d302253ac759392b2799bf2f5a32938745R161-R187).

Basically, you have to account for eras in which stake was 0.

E.g.:
current_era = 10;
 `[<1, 1000>, <2, 1500>, <3, 0>, <7, 1300>]` means that you can claim eras:
**1, 2, 7, 8, 9**

src/hooks/helper/index.ts
```diff
@@ -0,0 +1,4 @@
+import { SubmittableExtrinsic, SubmittableExtrinsicFunction } from '@polkadot/api/types';
+import { ISubmittableResult } from '@polkadot/types/types';
+
+export type BatchTxs = SubmittableExtrinsic<'promise', ISubmittableResult>[];
```
hoonsubin, line 4: I see that you try to make the type `BatchTxs` as an array of `SubmittableExtrinsic`, but please note that a batch transaction when it's submitted to the chain is considered as a single `SubmittableExtrinsic` with the input of `SubmittableExtrinsic[]`, this might be confusing for others (see https://github.com/hoonsubin/substrate-batch-script/blob/main/src/api/SubstrateApi.ts#L113)

src/hooks/useTvl.ts
```diff
@@ -27,8 +29,15 @@ export function useTvl(api: any) {
 
       const getTvl = async (): Promise<{ tvl: BN; tvlDefaultUnit: number }> => {
         const era = await apiRef.query.dappsStaking.currentEra();
-        const result = await apiRef.query.dappsStaking.eraRewardsAndStakes(era);
-        const tvl = result.unwrap().staked.valueOf();
+        let tvl;
+        if (isEnableIndividualClaim) {
```
hoonsubin, line 33: I feel like this can be made shorter. The code doesn't look clean

src/hooks/helper/claim.ts
impelcrypto: Sorry, I don't get it. `ContractEraStake` interface is used for querying `contractEraStake`.
image

src/hooks/helper/claim.ts
impelcrypto: Thank you, @Dinonard.
I fixed this issue in [this](https://github.com/AstarNetwork/astar-apps/pull/258/commits/77fbfd9ac61aef0a6a1645f9aa342e6b84d1cdb9) commit.

Demo: https://www.loom.com/share/d6fcde392cd94c14b3951cc84d39398b

Screenshot 2022-02-28 at 1 06 30 PM

Screenshot 2022-02-28 at 1 09 12 PM

src/hooks/helper/index.ts
impelcrypto: [Fixed](https://github.com/AstarNetwork/astar-apps/pull/258/commits/1e7f40e568c4c5b1162767efcb80cf8cb0436109)

src/hooks/useTvl.ts
impelcrypto: [Fixed](https://github.com/AstarNetwork/astar-apps/pull/258/commits/ebd8c28a4c863e721825685de02a7485f3d7ce2d)

src/config/chainEndpoints.ts
hoonsubin: This method of checking for available functions in the chain doesn't seem scalable. Please refactor this

impelcrypto: > This method of checking for available functions in the chain doesn't seem scalable. Please refactor this

@Dinonard Could you share with me how can I query the node version that is available for the individual claim feature, please?

impelcrypto:
@bobo-k2 
Could you advise on this please?
https://github.com/AstarNetwork/astar-apps/pull/258#issuecomment-1046534222

bobo-k2:
> @bobo-k2 May I know if it is ok to omit `nonce: -1`?
> 
> https://github.com/AstarNetwork/astar-apps/blob/73d0eac99c14beec5c7a1a2f84eb8f4d6e5d233d/src/store/dapp-staking/actions.ts#L551
> 
> https://github.com/AstarNetwork/astar-apps/blob/73d0eac99c14beec5c7a1a2f84eb8f4d6e5d233d/src/hooks/useIndividualClaim.ts#L60



> @bobo-k2 May I know if it is ok to omit `nonce: -1`?
> 
> https://github.com/AstarNetwork/astar-apps/blob/73d0eac99c14beec5c7a1a2f84eb8f4d6e5d233d/src/store/dapp-staking/actions.ts#L551
> 
> https://github.com/AstarNetwork/astar-apps/blob/73d0eac99c14beec5c7a1a2f84eb8f4d6e5d233d/src/hooks/useIndividualClaim.ts#L60
Please leave the nonce, because removing it can cause subsequent transactions to fail

src/config/chainEndpoints.ts
impelcrypto: Hi @Dinonard 
> V2_0_0 is the current one with unbonding period.

Is the way to check which node version is available for `individual claim` feature? Seems like both and Shiden the local node returns `v2_0_0`. 
image

Dinonard: Oh, I see.

The version will be changed to `V3_0_0` when we do runtime upgrade.
If you run local node, it will still say `V2_0_0`.

But let me change this so it's easier for you to test.
I will push an update so default version is 3.

impelcrypto: Thank you, @Dinonard 
btw, I found that there is no 'storageVersion' query in Astar network.

image

impelcrypto:
Hi @bobo-k2 

Could you help fix [getClaimInfo](https://github.com/AstarNetwork/astar-apps/blob/feat/individual-claim/src/store/dapp-staking/actions.ts#L652) function in this branch, please?🙏 I think this branch has filed to pull the main branch correctly because we changed the logic in this function.

Localhost (Shiden)
image

Production (shiden)
image

src/config/chainEndpoints.ts
impelcrypto: > This method of checking for available functions in the chain doesn't seem scalable. Please refactor this

@hoonsubin @bobo-k2 
[Fixed](https://github.com/AstarNetwork/astar-apps/pull/258/commits/a17dff706a57468b6c9bdf52654e0cf98efe96fb)

Dinonard: Astar still uses legacy runtime where no `version` storage exists - you can assume for now that it's 1.

Once we upgrade Astar, it will immediately become v3.

impelcrypto: > Once we upgrade Astar, it will immediately become v3.

@Dinonard 
I see. Am I correct that the v3 for Astar means the network is available for the `individual claim` feature?

Dinonard: `V3_0_0` means Individual claim is available, regardless of network.

In case no `Version` storage exists, this means that legacy dapps-staking is being used and the assumed version should be `V1_0_0`.

Hope this helps.

bobo-k2:
I am running a local node which has implemented individual claim. I built the node from https://github.com/AstarNetwork/Astar/tree/feature/individual-claim-local

impelcrypto:
> When I try to Claim all I got

@bobo-k2 
Did you stake on the contract address `0x0000000000000000000000000000000000000005`? Because the local node has 2 dApps registered under the same address above. Please try `0x0000000000000000000000000000000000000006`(Flipper) instead.
image

impelcrypto:
>In case of invalid address I will be unable to stake. 

Yes, you can stake but that address (...05) has been registered under 2 dApps that's why the error has occurred on the frontend. We shouldn't have this case in Mainnet (the same address registered on multiple dApps).

impelcrypto:
>  I registered 2nd contract 0x0000000000000000000000000000000000000006, stake on the contract and try to claim, but still getting the same error.

I'll try to test on the latest code of the [node](https://github.com/AstarNetwork/Astar/tree/feature/individual-claim-local) branch.

bobo-k2:
> Hi @bobo-k2 I have tested with the latest code of [feature/individual-claim-local]
I can confirm that everything works if I use contract ...06 only.

Dinonard:
@impelcrypto
I did the following:
* start clean chain `astar-collator --dev --tmp`
* registered a dapp using address 0x5 using Polkadot.js with Alice account
* staked on this dapp via portal (running locally)
* when next era started and rewards became claimable, this was displayed:
  * `Claim All (2)`.

Not sure where the **2** comes from or what it represents.

Dinonard:
@impelcrypto one additional point :)

I like the single `Claim all` button :+1: 

However, do we need to keep `Claim` buttons on separate contracts then?
It's confusing and will make UX bad for our users.

What do you think?

Dinonard:
@impelcrypto one more :smile: 

It's good you removed **Pending rewards** calculation and display, but we should display how much claimer earned.
This is important to show to the user - otherwise they just call claim and - nothing happens.
They would have to go to subscan or some explorer to see how much they earned.

This is important for the UX.

impelcrypto:
Hi @Dinonard 

> Claim All (2)

It means 2 dapps are claimable by the staker. However, our DB has 2 dApps registered under `0x5` that why it caused an error when you click `claimAll`. You can stake on `0x6` to avoid this issue.

https://github.com/AstarNetwork/astar-apps/pull/258#issuecomment-1060812601
https://github.com/AstarNetwork/astar-apps/pull/258#issuecomment-1060839674

impelcrypto:
>  but we should display how much claimer earned

@Dinonard 
Sure I'll check it tomorrow, but if I remember correctly, there is no way to determine which events belong to claimDapp or claimStaker in the emitted events.

impelcrypto:
> However, do we need to keep Claim buttons on separate contracts then?
It's confusing and will make UX bad for our users.
What do you think?

@hoonsubin 
What do you think about it? Perhaps remove the 'Claim' button or change it to 'Details'?

image

Dinonard:
> > but we should display how much claimer earned
> 
> @Dinonard Sure I'll check it tomorrow, but if I remember correctly, there is no way to determine which events belong to claimDapp or claimStaker in the emitted events.

Actually, you can :)

If you e.g. call claim_staker 4 times and then claim_dapp 4 times, the first 4 reward events will be for staker reward and the last 4 events will be for dapp reward.

Initially, it's enough just to show the user how much they've earned, imo.
We can create an additional ticket to improve this so `staker` and `dapp` reward are shown separately.

impelcrypto:
> but we should display how much claimer earned

@Dinonard 
I've added a toast :)

https://user-images.githubusercontent.com/92044428/157406514-fd538ee6-1d3c-49d8-8570-cc704e491d96.mov

image

Dinonard:
> > but we should display how much claimer earned
> 
> @Dinonard I've added a toast :)
> 
> image
> claim.toast.mov

Very nice! :partying_face:

impelcrypto:
Hi @hoonsubin 
Do you have any comments about this?

https://github.com/AstarNetwork/astar-apps/pull/258#issuecomment-1061760594
https://github.com/AstarNetwork/astar-apps/pull/258#issuecomment-1061813936

src/boot/api.ts
```diff
@@ -14,6 +18,7 @@ import Web3 from 'web3';
 
 const $api = ref<ApiPromise>();
 const $web3 = ref<Web3>();
+const $isEnableIndividualClaim = ref<boolean>(false);
```
hoonsubin, line 21: We have a module function check function https://github.com/AstarNetwork/astar-apps/blob/13e951fe5214a485448848eb1f4fb002ca6a7b39/src/hooks/useStakingH160.ts#L196 from the MetaMask dApps Staking PR. Having the module check spread around the code will make things difficult to maintain.

hoonsubin:
> @impelcrypto one additional point :)
> 
> I like the single `Claim all` button 👍
> 
> However, do we need to keep `Claim` buttons on separate contracts then? It's confusing and will make UX bad for our users.
> 
> What do you think?

I think this is more of a scope for the portal v2, no need to consider that this as long as the function works

hoonsubin:
> > However, do we need to keep Claim buttons on separate contracts then?
> > It's confusing and will make UX bad for our users.
> > What do you think?
> 
> @hoonsubin What do you think about it? Perhaps remove the 'Claim' button or change it to 'Details'?
> 
> image

I guess we can remove the claim button from the modal and change it to details, but the claim button on the item can simply claim to not confuse anyone.

hoonsubin (approved):
LGTM.
We will merge this when the chain side is ready

bobo-k2 (approved):
LGTM

PR #279, opened 2022-03-12
feat: added account component

impelcrypto:
**Pull Request Summary**

* Added /assets page
* Added `Account` component in /assets page 
* Hid /balance tab from sidebar
* Set up SASS configuration and the global variables (Applied the css class name by [BEM](https://css-tricks.com/bem-101/) methodology)
* Added `SF Pro` font

 
**Check list**
- [ ] contains breaking changes
- [x] adds new feature
- [ ] modifies existing feature (bug fix or improvements)
- [ ] relies on other tasks
- [ ] documentation changes

**To-dos**
* Redirect to /dashboard page whenever browser has not installed any wallet extensions
* Calculate the total amount (USD) of the assets

375px
image

1280px
image

1440~px
image

image

Gif:
https://gyazo.com/d163b5d21c7500a91b1b71e50bcd743d

When users haven't connected wallets yet
image

CC: @Kahonnohak

PR #282, opened 2022-03-16
feat: added assets component

impelcrypto:
**Pull Request Summary**

* Added `Native asset list` and `EVM asset list`
* Added `Account (Native)` for lockdrop account
* Changed text font to `Inter`

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
image

image

image

image


src/components/assets/Assets.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <div class="wrapper--assets">
+    <div v-if="!selectedAddress" class="backdrop--transparent" />
+    <div class="container--assets">
+      <Account :ttl-erc20-amount="ttlErc20Amount" />
+      <div v-if="!isLoading && selectedAddress">
```
hoonsubin, line 6: Screen Shot 2022-03-21 at 9 32 51 AM

After logging in, it takes time for the asset panel to show up.
I think it's better to add a loading skeleton for the contents

src/components/assets/Assets.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <div class="wrapper--assets">
+    <div v-if="!selectedAddress" class="backdrop--transparent" />
```
hoonsubin, line 3: Screen Shot 2022-03-21 at 9 32 22 AM

When I first access the page, the assets are blurred, and there are no further instructions from the page. What about opening the wallet connect modal when the user accesses the page?

src/components/assets/Account.vue
```diff
@@ -0,0 +1,236 @@
...
+        msg: 'Copy address success!!',
+        alertType: 'success',
+      });
+    };
+
+    watchEffect(() => {
+      const account = getSelectedAccount(substrateAccounts.value);
+      if (!currentAccount.value) {
+        // Memo: placeholder
+        iconWallet.value = supportWalletObj[SupportWallet.PolkadotJs].img;
```
hoonsubin, line 146: Instead of using the polkadot-js wallet icon? How about showing an identicon? https://github.com/polkadot-js/ui/tree/master/packages/vue-identicon
The polkadot-js is a wallet-specific logo.

cc @Kahonnohak

hoonsubin (changes requested):
Also, I'm getting an error message
```
invalid address (argument="address", value="Wh2nf6F5ZNJguoQu22Z361xo6VFqX1Y2BuQMcJBSJxERh5E", code=INVALID_ARGUMENT, version=address/5.5.0) (argument="_owner", value="Wh2nf6F5ZNJguoQu22Z361xo6VFqX1Y2BuQMcJBSJxERh5E", code=INVALID_ARGUMENT, version=abi/5.0.7)
```

src/components/assets/Assets.vue
hoonsubin: I also want to know my transferrable balance

src/components/assets/Assets.vue
impelcrypto: In that case, the sidebar will be opened automatically. The logic will be implemented once the `select wallet` component (global) has been implemented by @sirius651 .

https://gyazo.com/9dd2e3ea8f58875a7cbb7e81a43287c2

src/components/assets/Assets.vue
impelcrypto: > I also want to know my transferrable balance
@Kahonnohak :)

src/components/assets/Account.vue
impelcrypto: @hoonsubin 
It's not only polkadot.js but it displays each wallet icon here. Do you suggest replacing the wallet icon with the logo provided in [this](https://github.com/polkadot-js/ui/tree/master/packages/vue-identicon) package?

image

hoonsubin: I personally think using the identicon is better since all Polkadot web wallets use it. Plus, this method will be difficult for Ethereum wallet logins because users can choose to login with one wallet and actually sign the transaction with another (ex, Trust vs MetaMask). But I think I'll leave the final decision to @Kahonnohak since I like either options.

hoonsubin:
The head branch is set to `main`, but I think you might want to change this to `v2-development`

impelcrypto:
> The head branch is set to main, but I think you might want to change this to v2-development

You are right. Thank you, @hoonsubin

src/components/assets/Assets.vue
Kahonnohak: Yes, make sense. Total balance + transferrable or just transferrable might be fine. will get back to you this evening with other feedback thanks.

src/components/assets/Account.vue
Kahonnohak: That something I wanted to discuss thanks, my personal guess is no one is actually care about their identicon either js nor Metamask but this is something we can consider later on. I would say to show the wallet icon that people can really understand what wallet(account) you are connecting with. I'm not fully understand what you mention about Trust vs Metamask, would be great if we can discuss this bits at Dev sync later.

src/components/assets/styles/account.scss
```diff
@@ -1,5 +1,10 @@
 @import 'src/css/quasar.variables.scss';
 
+.wrapper--account {
```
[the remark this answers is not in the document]
hoonsubin: Previously, we used tailwindcss to quickly bootstrap a project without a dedicated designer, but since we now have someone, we can stop using tailwind from now on, especially if the project configuration is a hassle.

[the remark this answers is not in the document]
impelcrypto: I don't want to rely on tailwindcss too much as long as sass can handle the styling because I think styling and HTML should be not mixed. I want to make the `.vue` file as simple as possible.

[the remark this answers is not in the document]
hoonsubin: Then I think we can remove tailwind from the project. @sirius651 what do you think?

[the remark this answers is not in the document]
sirius651: agreed with simplicity of the components, but what if we just separate the contents of the class if too-much mixing is a problem? and if we go `scss`, it would be good to define the policy of `scss`. for example, what are the rules for making classNames like `row__right`, `column--balance`? how to maintain reusable global classes instead of boilerplate parts? In this point, `tailwind` can take care of it I think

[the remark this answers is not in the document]
impelcrypto: @sirius651 
I've applied the css class name by [BEM](https://css-tricks.com/bem-101/) methodology.
https://github.com/AstarNetwork/astar-apps/pull/279#issue-1167205769

src/components/assets/Assets.vue
Kahonnohak: スクリーンショット 2022-03-22 9 44 26
@impelcrypto @hoonsubin 
The balance at the top is total balance and the faucet would be here when the balance is zero.
The rest of the lines of the related balance up below. I think it is ok to show zero balance for all (don't have to hide) as then people know that you haven't got any otherwise they would ask. But also don't have to make a change if the current setting is to hide for zero balance.

Kahonnohak:
@impelcrypto For the first landing view, would it be possible to make blur widen and soften please? if it would take time to experiment I would go with plan B for not blur and show balance 0.


スクリーンショット 2022-03-22 10 56 51
スクリーンショット 2022-03-22 10 57 00

src/components/assets/Assets.vue
impelcrypto: @Kahonnohak Pushed, please take a look at the [staging](https://deploy-preview-pr-282.herokuapp.com/) url.

impelcrypto:
> For the first landing view, would it be possible to make blur widen and soften please? if it would take time to experiment I would go with plan B for not blur and show balance 0.

@Kahonnohak 
As spoken, let's improve this in the future.

src/components/assets/styles/account.scss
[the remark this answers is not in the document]
sirius651: I see, in that case, I also should follow that rule

Kahonnohak (approved):
It is amazing to be able to see all ASTR balance at once! Thanks these look great.

hoonsubin (approved):
LGTM

PR #333, opened 2022-04-06
Set fix version of ui library : v0.0.15

sirius651:
**Pull Request Summary**

**Check list**
- [ ] adds new feature
- [x] relies on other tasks
- [ ] documentation changes

I'll keep updating & publishing to astar-ui library. but before that, currently the portal brings the latest version of ui lib, so we should fix as specific version(v0.0.15) for the possibility of crashes or bugs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@astar-network/astar-api": "^0.1.5-3",
-    "@astar-network/astar-ui": "^0.0.15",
+    "@astar-network/astar-ui": "0.0.15",
     "@ethersproject/bignumber": "^5.5.0",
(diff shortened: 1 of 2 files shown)


impelcrypto (approved):

Kahonnohak (approved):

Kahonnohak:
@sirius651 Hello! I just approved the PR by mistake. Could you tell which UI you made change and need review please?

Kahonnohak:
Thanks!

In terms of what we would like to have in UI library - Based on my experience UI parts are useful when it is already scaled and need to produce 1000+ pages. However for instance atomic design is so difficult to maintain when you want to make a little change. Yes agree we should discuss what UI we need for the library.

PR #341, opened 2022-04-11
Compounding rewards

bobo-k2:
**Pull Request Summary**
Staging url
https://astar-apps--pr341-feature-compound-rew-9cxujnv7.web.app/#/dapp-staking/discover

This PR brings UI support for compounding rewards and dapps staking page layout improvements.
Your rewards panel is displayed only to users who have been staked

image

The PR to be merged after compounding rewards feature is deployed to Shibuya. 

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- compounding support

**Changes**
- UI updates


Kahonnohak:
@bobo-k2 I can't see the compounding feature and APR on Shibuya. Am I seeing how it should be?

スクリーンショット 2022-04-11 13 38 42

src/components/dapp-staking/statistics/CompoundReward.vue
```diff
@@ -0,0 +1,59 @@
...
+    <div class="title">
+      {{ $t('dappStaking.autoCompound') }}
+      <IconTooltip>
+        {{ $t('dappStaking.autoCompoundingTooltip') }}
+      </IconTooltip>
+    </div>
+    <div>
+      <span class="text--title">{{ isCompounding ? 'ON' : 'OFF' }}</span>
+      <Button :small="true" :primary="true" class="button" @click="changeDestination">
+        {{ isCompounding ? 'Turn off' : 'Turn on' }}
```
hoonsubin, line 12: Please use i18n

Dinonard (commented):
A few questions/comments:

1. Why does the panel look differently when compounding is turned off?
Screenshot from 2022-04-27 18-40-05

2. In case dApp is Unregistered or staker has fully unstaked (latest staked value is zero), compounding rewards won't work - the extrinsic won't fail but rewards won't be restaked.
Also, in case the `StakingInfo.stakes` would go above `MAX_ERA_STAKE_VALUES`, the call will fail since it would increase the vector size past the max size limit. User should then switch compounding rewards off before claiming.
Now, this won't happen often but it's worth mentioning that it's a possibility. We don't want someone calling `claim` while it fails all the time.
Maybe a task for later? Some improvement ticket?

bobo-k2:
> A few questions/comments:
1. Only difference I see is that Claim button is displayed. Is the Claim button what you are referring to?
2. I will create separate task for situations when dApp is Unregistered or staker has fully unstaked.
I am handling the situation when error received from setRewardDestination call contains TooManyEraStakeValues. In that case the message will be shown 'Disable compounding, claim your rewards and then enable compounding again.'

impelcrypto:
image

@bobo-k2 
Is the auto compound turned on by default?

impelcrypto (approved):

bobo-k2:
> @bobo-k2 Is the auto compound turned on by default?

As far as I know yes, but I am sure @Dinonard knows exact answer to this

Dinonard:
> > A few questions/comments:
> 
>     1. Only difference I see is that Claim button is displayed. Is the Claim button what you are referring to?
> 
>     2. I will create separate task for situations when dApp is Unregistered or staker has fully unstaked.
>        I am handling the situation when error received from setRewardDestination call contains TooManyEraStakeValues. In that case the error message will contain 'Disable compounding, claim your rewards and then enable compounding again.'

1. Yes, the `Claim` button is different. Just thought I'd mention it since I noticed it.

2. `setRewardDestination` won't throw `TooManyEraStakeValues`. Do you mean you check whether this condition is true if user tries to configure automatic restake?

Dinonard:
> > @bobo-k2 Is the auto compound turned on by default?
> 
> As far as I know yes, but I am sure @Dinonard knows exact answer to this

Yes, the DB value by default will be to compound.

bobo-k2:
> > > A few questions/comments:
> > 
> > 
> > ```
> > 1. Only difference I see is that Claim button is displayed. Is the Claim button what you are referring to?
> > 
> > 2. I will create separate task for situations when dApp is Unregistered or staker has fully unstaked.
> >    I am handling the situation when error received from setRewardDestination call contains TooManyEraStakeValues. In that case the error message will contain 'Disable compounding, claim your rewards and then enable compounding again.'
> > ```
> 
> 1. Yes, the `Claim` button is different. Just thought I'd mention it since I noticed it.
> 2. `setRewardDestination` won't throw `TooManyEraStakeValues`. Do you mean you check whether this condition is true if user tries to configure automatic restake?

1. In case when auto compounding is off we wanted to give user an option to claim rewards manually
2. I understood it will be exception. I will implement condition check

Kahonnohak:
A bit of styling adjustment please. I understand in this layout it won't fit with other cards but when it comes to responsiveness this is better, so it is ok.

スクリーンショット 2022-04-29 16 02 01 スクリーンショット 2022-04-29 15 55 01

Kahonnohak:
And for the tool tip.

Please adjust it so that it is centred with the icon. Colour change to match with the button.

And please check all texts in the tooltip, they need polish.

スクリーンショット 2022-04-29 16 16 51

Dinonard:
@bobo-k2 `TooManyEraStakeValues` error is raised by `claim_staker` call.

E.g. if current era is `10`, `MaxEraStakeValues = 3` and `StakingInfo.stakes = [(3, 100), (5, 150), (7, 120)]`

If `claim_staker` is called, vector of stakes becomes `[(4, 100), (5, 150), (7, 120), (10, 121)]`
Since new size equals `4`, `TooManyEraStakeValues` error is raised.

bobo-k2:
> Thank you @bobo-k2 Looks good!

@Kahonnohak please check now

src/components/dapp-staking/statistics/CompoundReward.vue
bobo-k2: > Please use i18n

@hoonsubin this has been fixed. Please check.

hoonsubin:
Screen Shot 2022-05-02 at 10 39 27 AM

How about being able to auto restake per-dapp? Right now, the option is global, but what if I want to restake for certain dapps and be able to claim rewards for some?

cc @Dinonard @Kahonnohak

Dinonard:
@hoonsubin It's a config per account, not per dApp in the backend.

We can consider this in the future if needed.
How many stakers stake on more than 1 dApp?

Anyhow, this shouldn't be a blocker and we should proceed with deployment as soon as possible.

hoonsubin:
> @hoonsubin It's a config per account, not per dApp in the backend.
> 
> We can consider this in the future if needed. How many stakers stake on more than 1 dApp?
> 
> Anyhow, this shouldn't be a blocker and we should proceed with deployment as soon as possible.

Understood. If we wish to make dApps Staking a developer incentivizing mechanism, it makes more sense to encourage people to stake on multiple dApps, but I will admit, this is more of a nice to have feature and not a must feature.

hoonsubin (approved):
LGTM

Dinonard:
@bobo-k2 I've just checked the Shibuya and I cannot see the `Claim` button when `AutoCompund` is ON.
When I turn it off, I can see the `Claim` button.

My `EraStake` vec only has 1 value.

bobo-k2:
> @bobo-k2 I've just checked the Shibuya and I cannot see the `Claim` button when `AutoCompund` is ON. When I turn it off, I can see the `Claim` button.
> 
> My `EraStake` vec only has 1 value.

@Dinonard this is by design from @Kahonnohak. If I understood correctly when auto-compund is on rewards are automatically staked so user doesn't need to claim

Kahonnohak:
@bobo-k2 Thanks so much!

@Dinonard 
I'm confused. What could you claim when rewards are automatically re-staked?

Dinonard:
@Kahonnohak I think we misunderstood each other how the feature works.

The _only_ difference when auto-compound is enabled is that when `claim_staker` call is used, it will payout the reward to the staker AND automatically re-stake that reward.

So interaction from the user is still required - it's not _set and forget_.
It would be very expensive for our network/protocol to automatically claim & re-stake rewards for all stakers when era ends.

Kahonnohak:
aha! I got it now. I believe in general people would probably think the way I was thinking.

Let's make it clear,

1) please change the order - Claim comes first, Auto-Compound comes after.

2) Auto-Compound to be "Auto-compound after claiming"

3) caption to be - "By turning on the Auto compound, your rewards will automatically re-stake after claiming. "

What would you think @Dinonard ? Please correct my texts if in correct.
スクリーンショット 2022-05-04 16 19 09
スクリーンショット 2022-05-04 16 19 22

Dinonard:
@Kahonnohak looks good! :+1: 

And sorry for the misunderstanding.

Kahonnohak:
Great! This should still help keep our TVL! Let them come back to the portal often and not let them forget XD
 @bobo-k2 Please see above and make small changes.

PR #347, opened 2022-04-13
Enable maintenance

bobo-k2:
**Pull Request Summary**

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)


impelcrypto (approved):

src/store/dapp-staking/actions.ts
```diff
@@ -798,15 +798,15 @@ const actions: ActionTree<State, StateInterface> = {
...
-        commit('setIsPalletDisabled', isPalletDisabled);
+        // let isPalletDisabled = false;
+        // try {
+        //   const isDisabled = await $api.value.query.dappsStaking.palletDisabled<bool>();
+        //   isPalletDisabled = isDisabled.valueOf();
+        // } catch {
+        //   // palletDisabled storage item is not supported by a node;
+        // }
+
+        commit('setIsPalletDisabled', true);
```
hoonsubin, line 809: Can you make it so that it can allow Shiden users to continue using dapps staking while only filter astar?

hoonsubin (changes requested):

fiexer (approved):

src/store/dapp-staking/actions.ts
impelcrypto: @bobo-k2 
I see there is no `palletDisabled` query in Astar. How about:
```
// Memo: "Astar" or allocate node version like `isIndividualClaim`
if(network === Astar){
  return true
}else{
  const isDisabled = await $api.value.query.dappsStaking.palletDisabled<bool>();
  return isDisabled.valueOf();
}
```

hoonsubin: Sure, that works

bobo-k2: I made a quick fix. I will merge now and polish later

bobo-k2: Added some improvements

hoonsubin (approved):
LGTM

PR #382, opened 2022-05-04
Compounding rewards layout and text updates

bobo-k2:
**Pull Request Summary**

Updates as per latest conversation in https://github.com/AstarNetwork/astar-apps/pull/341#issuecomment-1117007575

image

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- Components reordering and text updates

### DIFF
--- a/src/components/dapp-staking/DiscoverDappsTab.vue
+++ b/src/components/dapp-staking/DiscoverDappsTab.vue
@@ -18,48 +18,7 @@
       <DappsCount />
       <Era :progress="progress" :blocks-until-next-era="blocksUntilNextEra" :era="era" />
       <APR />
-      <!-- <Withdraw /> -->
     </div>
-
-    <!-- <div class="tw-text-center tw-mb-8 tw-flex tw-items-center tw-justify-center sm:tw-gap-x-4">
-      <Button :disabled="isPalletDisabled" @click="showRegisterDappModal = true">
-        <astar-icon-base
-          class="tw-w-5 tw-h-5 tw-text-white tw--ml-1"
-          stroke="currentColor"
-          icon-name="plus"
-        >
-          <astar-icon-plus />
-        </astar-icon-base>
-        {{ $t('dappStaking.registerDapp') }}
-      </Button>
-      <div
-        v-if="stakerApr > 0"
-        class="
-          sm:tw-w-40
-          tw-justify-center
-          tw-inline-flex
-          tw-items-center
-          tw-px-6
-          tw-py-3
-          tw-border
-          tw-border-transparent
-          tw-text-sm
-          tw-font-medium
-          tw-rounded-full
-          tw-shadow-sm
-          tw-text-white
-          tw-bg-indigo-500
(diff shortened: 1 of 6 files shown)


Dinonard:
@bobo-k2 I tried it out and it works great :+1:

impelcrypto:
I found that the `compound rewards panel` won't [update](https://github.com/AstarNetwork/astar-apps/blob/feature/compund-rewards-polishing/src/components/dapp-staking/UserRewards.vue#L2) whenever users change their accounts. Let's use watchEffect [here](https://github.com/AstarNetwork/astar-apps/blob/feature/compund-rewards-polishing/src/hooks/dapps-staking/useCompoundRewards.ts#L137).


Demo:
https://www.loom.com/share/26c5fb2684c34d81bc891d0efd1f4521

bobo-k2:
> I found that the `your rewards panel` is not going to [update]

Thanks @impelcrypto, fixed

impelcrypto (approved):

hoonsubin:
@Kahonnohak The panel seems a bit complex now... Instead of adding the Auto-compound toggle, what if we add another button next to the `Claim` saying `Restake` which will restake the claimed rewards?

Kahonnohak:
I get what you say, Auto-Compound here really is an additional option for Claim. "Re-staking rewards" sounds appropriate maybe? (But then you have to say "would like to re-stake the rewards?" to clarify which is a bit too long.  It has to have ON or OFF  status and the default is on. User shouldn't press the button if he would like to re-stake.  Hence I think we should go as it is and see how people use them.

Kahonnohak:
@hoonsubin Oh is this mean that Users have options
1) Claim - Claim means claim, the rewards will be added to Transferable balance.
2) Restake - User will need two to sign 2 transactions for Claim and Restake, which I think it is good. But @Dinonard will need backend work for Restake feature as this will be a button that shouldn't have the status of ON or OFF.  

What do you think?
スクリーンショット 2022-05-04 21 41 33

Dinonard:
@hoonsubin @Kahonnohak 
Backend shouldn't be changed for this because it would require a storage migration.
We discussed this before on some meeting and agreed that it should be set _once_ by user and then it should apply to all claims. This is in line how polkadot/kusama staking works.

The plan to launch this feature on Shiden is tomorrow - imo, the final version of the UI should have been already deployed and live for Shibuya.

I agree with Kahori that we should go with what we have and look into improving it for v2 portal.

Kahonnohak:
I'm happy to approve as it is but will wait for the re-stake behaviour to be cleared.

Dinonard:
What could be done is a prompt for the user that would act _as if_ it functions like @hoonsubin thought it does.
Then if user isn't happy with the selected option, they change it and at the beginning of a batch call, reward destination change is inserted.

But right now we need this current solution deployed :sweat_smile:

Kahonnohak (approved):
LGTM:)

PR #383, opened 2022-05-05
Claim all disabled when claimable awards available

bobo-k2:
**Pull Request Summary**

This bug was caused because era number was converted to string and handled as such in getNumberOfUnclaimedEra.
When era was above 1000 it was formatted like '1,000' which caused math operations to return NaN.

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


hoonsubin (approved):
LGTM
I already reviewed it but forgot to approve lol

impelcrypto (approved):

PR #386, opened 2022-05-06
fix: connecting wallet using firefox

impelcrypto:
**Pull Request Summary**

* fix: connecting wallet using Firefox 

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Fixes**

Firefox takes some time to load the wallet extensions at the boot time.

=Before=

image

=After=

image

### DIFF
--- a/src/config/api/polkadot/connectApi.ts
+++ b/src/config/api/polkadot/connectApi.ts
@@ -7,7 +7,7 @@ import { isTestChain } from '@polkadot/util';
 import { cryptoWaitReady } from '@polkadot/util-crypto';
 import { endpointKey, providerEndpoints } from 'src/config/chainEndpoints';
 import { LOCAL_STORAGE } from 'src/config/localStorage';
-import { objToArray } from 'src/hooks/helper/common';
+import { objToArray, wait } from 'src/hooks/helper/common';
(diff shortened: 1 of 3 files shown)


src/config/api/polkadot/connectApi.ts
```diff
@@ -160,6 +160,8 @@ export async function connectApi(
     fallbackConnection({ networkIdx, endpoint });
   }
 
+  // Memo: wait for the wallet extensions to load in Firefox
+  await wait(400);
```
bobo-k2, line 164: We can't be sure that extension will be loaded in 400ms. Is there a event we can listen or similar?

bobo-k2 (changes requested):

src/config/api/polkadot/connectApi.ts
impelcrypto: You are right. How about [this](https://github.com/AstarNetwork/astar-apps/pull/386/commits/8e43b9f68f04079709c552763b31b92a362c0146#diff-f3c7b24fa7deac4c9c27bae263642406de5d5e7e4abc997af4afae4f24e04109R17) solution?

hoonsubin (approved):
LGTM

PR #400, opened 2022-05-17
feat: updated XCM bridge modal

impelcrypto:
**Pull Request Summary**

* updated the XCM bridge modal styling
* improved the connection to the relaychain API

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Fixes**

* feat: update the XCM asset balances after sending bridge transaction
* feat: improved the connection to the relaychain API
  * resolved errors that came from SubstrateApi.ts
  * improved fetching speed for the account balances in relaychain
    * before: around 10secs
    * after: around 5 secs
* styling: updated the modal styling 
* styling: display loader modal while loading the account balances and existential deposit amount 
* styling: close XCM bridge modal after sending bridge transaction

=Before=
 errors that came from SubstrateApi.ts
image

image

=After=
image

image

### DIFF
--- a/src/components/assets/modals/ModalXcmBridge.vue
+++ b/src/components/assets/modals/ModalXcmBridge.vue
@@ -43,7 +43,7 @@
             <div />
           </div>
           <div class="row__chain">
-            <img :src="chainIcon.dest" alt="src-chain-logo" class="logo" />
+            <img :src="chainIcon.dest" alt="dest-chain-logo" class="logo" />
             <span class="text--title"> {{ chainName.dest }} </span>
           </div>
         </div>
--- a/src/components/assets/modals/ModalXcmBridge.vue
+++ b/src/components/assets/modals/ModalXcmBridge.vue
@@ -34,7 +34,7 @@
           </div>
           <div class="row__chain">
             <img :src="chainIcon.src" alt="src-chain-logo" class="logo" />
-            <span class="text--xl"> {{ chainName.src }} </span>
+            <span class="text--title"> {{ chainName.src }} </span>
           </div>
         </div>
         <div class="box--input-chain">
@@ -44,7 +44,7 @@
           </div>
           <div class="row__chain">
             <img :src="chainIcon.dest" alt="src-chain-logo" class="logo" />
-            <span class="text--xl"> {{ chainName.dest }} </span>
+            <span class="text--title"> {{ chainName.dest }} </span>
           </div>
         </div>
 
--- a/src/components/assets/modals/ModalXcmBridge.vue
+++ b/src/components/assets/modals/ModalXcmBridge.vue
@@ -132,7 +132,7 @@
         </div>
       </div>
       <div class="wrapper__row--button">
-        <button class="btn btn--confirm" :disabled="isDisabledBridge" @click="bridge">
+        <button class="btn btn--confirm" :disabled="isDisabledBridge" @click="handleBridge">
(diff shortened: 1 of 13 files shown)


src/components/assets/modals/ModalXcmBridge.vue
```diff
@@ -98,6 +99,38 @@
...
+                }}
+              </span>
+            </div>
+          </q-tooltip>
+        </div>
+        <div class="row--warning">
+          <div class="column--title">
+            <span class="text--dot">・</span>
+            <span class="text--warning">
+              {{ $t('assets.modals.xcmWarning.fee') }}
```
hoonsubin, line 129: How about adding the actual fee for the transaction instead of a warning message?

src/hooks/xcm/SubstrateApi.ts
```diff
@@ -116,13 +116,22 @@ class ChainApi {
...
     try {
+      await this._api?.isReady;
       return ((await this._api.query.system.account(address)) as any).data.free.toBn() as BN;
     } catch (e) {
       console.error(e);
       return new BN(0);
     }
   }
 
+  public async isReady(): Promise<void> {
```
hoonsubin, line 127: Screen Shot 2022-05-17 at 2 09 48 PM

I tried to transfer right after I XCMed DOT, but I'm getting an RPC connection error. It seems like the portal failed to reconnect to Astar Network after finishing the XCM transfer from the relaychain

hoonsubin (changes requested):

src/components/assets/modals/ModalXcmBridge.vue
Kahonnohak: I thought that would take some time. That's good!

hoonsubin: I think you can do it with

```
const data = await tx.paymentInfo(signer.pair)
data.partialFee;
```

hoonsubin (changes requested):

src/components/assets/modals/ModalXcmBridge.vue
impelcrypto: I think `data.partialFee` returns the fee of the sending transaction. But we might want to display the deducted amount from receiving account. 
image

hoonsubin (approved):
LGTM

src/components/assets/modals/ModalXcmBridge.vue
hoonsubin: that should be the only fee there is I think? The only other method I know is https://github.com/polkadot-js/api/blob/1404994792db92f82bdeb66df85a372fbbc4ed1d/packages/rpc-augment/src/augment/jsonrpc.ts#L393 but this does not work when I tried

src/hooks/xcm/SubstrateApi.ts
impelcrypto: Transfer feature has not implemented yet. @bobo-k2 is working on it.

src/components/assets/modals/ModalXcmBridge.vue
impelcrypto: > that should be the only fee there is I think?

My EVM account received only 0.0006 DOT after sending 0.001 DOT from the native account
https://www.loom.com/share/3a5a9207396d462ab6c186d5e152f4f4

PR #403, opened 2022-05-18
Check if the non-zero balance of the account when transfer

sirius651:
**Pull Request Summary**

### DIFF
--- a/src/hooks/xcm/useXcmBridge.ts
+++ b/src/hooks/xcm/useXcmBridge.ts
@@ -357,6 +357,19 @@ export function useXcmBridge(selectedToken?: Ref<ChainAsset>) {
         return;
       }
 
+      // check if recipient account has non-zero native asset. (it cannot be transferred to an account with 0 nonce)
+      if (isValidEvmAddress(toAddress)) {
+        const balWei = await getBalance($web3.value!, toAddress);
+        if (Number(ethers.utils.formatEther(balWei)) === 0) {
+          throw Error('the balance of recipient account should be above zero');
+        }
+      } else {
+        const balData = ((await $api!.query.system.account(toAddress)) as any).data;
+        if (balData.free.toBn().eqn(0)) {
+          throw Error('the balance of recipient account should be above zero');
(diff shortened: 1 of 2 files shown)


src/hooks/xcm/useXcmBridge.ts
```diff
@@ -357,6 +357,19 @@ export function useXcmBridge(selectedToken?: Ref<ChainAsset>) {
         return;
       }
 
+      // check if recipient account has non-zero native asset. (it cannot be transferred to an account with 0 nonce)
+      if (isValidEvmAddress(toAddress)) {
+        const balWei = await getBalance($web3.value!, toAddress);
+        if (Number(ethers.utils.formatEther(balWei)) === 0) {
+          throw Error('the balance of recipient account should be above zero');
```
bobo-k2, line 364: Please put this into i18n file since the text is repeating

bobo-k2 (changes requested):

Kahonnohak:
Checked and it failed as it should be thanks!

Just to note, XCM has different UI - when Portal knows the receiver has less than required existential native token deposit, the transfer button is disabled. Here instead the button works and immediately fails. Fine for now. important that it fails.

impelcrypto (approved):

src/hooks/xcm/useXcmBridge.ts
sirius651: no idea how to use i18n in .ts file other than .vue. maybe in later, can tackle this. so refactored as separate variable for repeating words.

bobo-k2: you can do something like.
```
const { t } = useI18n();

customMsg.value = t('dappStaking.toast.staked', {
            amount: stakeAmount,
            dapp: props.dapp.name,
});
```

Check StakePanel.vue for full implementation

sirius651: Fixed

bobo-k2 (approved):
LGTM

PR #435, opened 2022-06-06
New architecture - get and display TVL

bobo-k2:
**Pull Request Summary**

Integrated the new architecture into the portal and implemented logic for getting dapps staking TVL.

Please check and comment. 

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- new portal architecture proposal


impelcrypto (approved):

src/v2/services/implementations/DappStakingService.ts
```diff
@@ -0,0 +1,35 @@
...
+import { IDappStakingService } from 'src/v2/services';
+
+@injectable()
+export class DappStakingService implements IDappStakingService {
+  constructor(
+    @inject() private dappStakingRepository: IDappStakingRepository,
+    @inject(Symbols.CoinGecko) private priceRepository: IPriceRepository,
+    @inject() private metadataRepository: IMetadataRepository
+  ) {
+    Guard.ThrowIfUndefined('dappStakingRepository', dappStakingRepository);
```
sirius651, line 16: Overall codes are great. but just curious that do we need to always put this kind of function to check undefined for all injectable params? it seems boilerplate a bit.

bobo-k2: @sirius651 Thank you for the review. I believe you are right. We don't need to check injected parameters since TypeScript wont' allow undefined values there. I will update that. On the other hand we need to check function parameters (e.g. strings) which will help us to notice/prevent bugs earlier.

bobo-k2: Fixed in commit https://github.com/AstarNetwork/astar-apps/pull/435/commits/8ae4688b257b28deb7f37958713e6b6d9920da53

sirius651 (approved):
LGTM

hoonsubin (approved):

src/v2/repositories/implementations/CoinGeckoPriceRepository.ts
```diff
@@ -0,0 +1,21 @@
+import axios from 'axios';
+import { Guard } from 'src/v2//common';
+import { IPriceRepository } from 'src/v2/repositories';
+import { injectable } from 'inversify-props';
+
+@injectable()
+export class CoinGeckoPriceRepository implements IPriceRepository {
```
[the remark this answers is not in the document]
impelcrypto: Also, I think `ticker` is better than `token id` (the id defined by coingecko) that is used for the function parameter since we might re-use this class to fetch the price for other ERC20 tokens. So the app doesn't need to know the coingecko id to fetch the price.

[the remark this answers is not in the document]
bobo-k2: Thank for the comment @impelcrypto I will implement DIA repository and failover in next iteration. Also I will change to ticker.

PR #440, opened 2022-06-11
New architecture sign transaction

bobo-k2:
**Pull Request Summary**

Implemented Stake method in `DappStakingService` class
image

Implemented `PolkadtoWalletService` and `MetamaskWalletService`. Wallet instance to be used is based on user wallet selection. Handled in `App.vue`
image

Implemented `EventAggregator` which is message broker. I decided to use messaging because it was easier to decouple current portal and v2 implementation. At the moment aggregator is used to provide transaction status to the portal so busy indicator or toast message are displayed properly. Message handlers are registered in `App.vue`
image

Also, the broker is used to provide gas price from `useGasPrice` hook to `GasPriceProvider` class.
image

Finally, `StakePanel.vue` component is modified to use new `DappStakingService` implementation.
image
 

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- Wallet components and message broker (EventAggregator)

**Changes**
- stake method moved to the new architecture

**To-dos**
- more testing


hoonsubin:
unsure if this is related to the changes you made, but I got the following error when I try to access the page.
Screen Shot 2022-06-13 at 8 28 46 AM
The portal will be completely white and won't display anything. This does not happen on production

bobo-k2:
The error is not happening on my local environment. I will take a look.

bobo-k2:
> unsure if this is related to the changes you made, but I got the following error when I try to access the page. 

This has been fixed. Please refresh staging env in browser with "Empty cache and hard reload"

impelcrypto (approved):

sirius651 (approved):

src/App.vue
```diff
@@ -31,13 +31,18 @@
 <script lang="ts">
 // Fix for breaking change introduced in polkadot js v7.x
 // https://polkadot.js.org/docs/api/FAQ/#since-upgrading-to-the-7x-series-typescript-augmentation-is-missing
+import 'reflect-metadata';
```
hoonsubin, line 34: How about moving this and `import '@polkadot/api-augment';` to `src/index.ts` since it's used globally?

hoonsubin (approved):

src/App.vue
bobo-k2: I don't see src/index.ts. Quasar is using multiple boot files to start the portal. 
image
Maybe I can create a new boot file an put `import 'reflect-metadata'` there

hoonsubin: hmm, actually no. I thought there was a index.ts root file (like most programs), but I must have been mistaken.
Since App.vue is the root file, let's just keep it here.

Can I merge this PR?

PR #451, opened 2022-06-22
fix: tooltip trigger logic for mobile devices

impelcrypto:
**Pull Request Summary**

* fix: changed tooltip trigger logic for mobile devices

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes
- [x] tested on mobile devices

**Fixes**
* fix: changed tooltip trigger logic for mobile devices
  * ref:  https://www.notion.so/astarnetwork/Mobile-tooltip-to-be-usable-ccc245bc1bab4adeb1f7d24ba89f88e8
  * this PR won't change the UX for PC users

=Before=
The tooltip works only when users are tapping the icon on mobile devices.
https://gyazo.com/d299af7deec3c6de12a9f896b0621d2d

=After=
An opened tooltip will be closed after users tap the outside of the icon.

https://gyazo.com/b83f2e976b7759343cd43a4c77dbabc6

https://gyazo.com/abeeccf399aeee932a5deec0d95ecec8


Kahonnohak:
Thank you! I can now see it in Dev tool!

- I think we can make the tooltip 12px. People can still read. 
- Texts align to the left, please.

Can we make the width narrower please?  We probably need the wider one for the next tooltip (longer) 

Can we make a rule:
- the width is 184px until texts are up to four lines
 -A after 4lines the box gets wider

https://www.figma.com/proto/NkD5tyIAQmbl6donW4eZma/Portal-V1-Kaizen?page-id=0%3A1&node-id=16%3A943&viewport=-1934%2C430%2C0.25&scaling=min-zoom

impelcrypto:
@Kahonnohak  Updated as per your feedback. Please take a look again.

Kahonnohak (approved):
Thank you so much! LGTM!

src/components/assets/modals/ModalXcmBridge.vue
```diff
@@ -172,8 +183,24 @@ export default defineComponent({
   },
   setup(props) {
     const isClosingModal = ref<boolean>(false);
+    const isMobileDisplayTooltip = ref<boolean>(false);
     const token = computed(() => props.token);
 
+    const isDisplayTooltip = computed<boolean | null>(() => {
```
bobo-k2, line 189: I noticed this code block is repeating 3 times in different files. Can you refactor so the logic is defined only once?

bobo-k2 (changes requested):

src/components/assets/modals/ModalXcmBridge.vue
impelcrypto: Sure, I will create a custom hook then.

impelcrypto: @bobo-k2 
Fixed in [this](https://github.com/AstarNetwork/astar-apps/pull/451/commits/29d0b5b531e241f02561932129c2a799ade187d2) commit.

hoonsubin (approved):

src/components/assets/modals/ModalXcmBridge.vue
impelcrypto: @bobo-k2 Please help take a look my comments.

bobo-k2 (approved):
LGTM

PR #476, opened 2022-07-29
fix: modified the position of the loading overlay

impelcrypto:
**Pull Request Summary**

* fix: modified the position of the loading overlay so that users can change the network endpoint while the page is loading

**Check list**
- [ ] contains breaking changes
- [ ] adds new feature
- [x] modifies existing feature (bug fix or improvements)
- [ ] relies on other tasks
- [ ] documentation changes
- [x] tested on mobile devices

**This pull request makes the following changes:**

**Changes**

Screenshot 2022-07-29 at 5 14 56 PM

PR #477, opened 2022-07-30
fix: modified z-index order in modals

impelcrypto:
**Pull Request Summary**

* fix: modified modal's z-index order
  * related: https://github.com/AstarNetwork/astar-apps/pull/476
* fix: modified styling in 'Transferable balance' for mobile devices
* fix: existential deposit check in XCM transaction

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Fixes**
* fix: modified modal's z-index order
=Before=
image

=After=
image

* fix: modified styling in 'Transferable balance' for mobile devices
=Before=
image

=After=
image

* fix: existential deposit check in XCM transaction
=Before=
image

=After=
image

### DIFF
--- a/src/components/assets/Assets.vue
+++ b/src/components/assets/Assets.vue
@@ -27,25 +27,21 @@
         </div>
       </div>
     </div>
-    <Teleport to="#app--main">
-      <div :class="!isLoading && 'highest-z-index'">
-        <ModalXcmTransfer
-          :is-modal-xcm-transfer="isModalXcmTransfer"
-          :handle-modal-xcm-transfer="handleModalXcmTransfer"
-          :handle-update-xcm-token-balances="handleUpdateXcmTokenAssets"
-          :account-data="accountData"
-          :token="token"
-        />
+    <ModalXcmTransfer
+      :is-modal-xcm-transfer="isModalXcmTransfer"
+      :handle-modal-xcm-transfer="handleModalXcmTransfer"
+      :handle-update-xcm-token-balances="handleUpdateXcmTokenAssets"
+      :account-data="accountData"
+      :token="token"
+    />
 
-        <ModalXcmBridge
-          :is-modal-xcm-bridge="isModalXcmBridge"
-          :handle-modal-xcm-bridge="handleModalXcmBridge"
-          :account-data="accountData"
-          :token="token"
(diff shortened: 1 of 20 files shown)


src/components/assets/Assets.vue
```diff
@@ -28,21 +28,23 @@
...
-      />
-
-      <ModalXcmBridge
-        :is-modal-xcm-bridge="isModalXcmBridge"
-        :handle-modal-xcm-bridge="handleModalXcmBridge"
-        :account-data="accountData"
-        :token="token"
-        :handle-update-xcm-token-balances="handleUpdateXcmTokenAssets"
-      />
+      <div :class="!isLoading && 'highest-z-index'">
```
hoonsubin, line 31: Correct me if I'm wrong, but it seems easier to change the modal style to have the highest z index rather than manually coding this for all components.

impelcrypto: It is so tricky because we have to consider the `z-index` of the three elements, `Header`, `Modals`, and `Loading animation` under different situations. The `z-index` priority must change dynamically.

We cannot make `Modals` always located at the highest position because the `Loading animation` will be hidden by modals. We also cannot make the `Loading animation` at the highest position because we might want to change the network when the APP is connecting to an endpoint. We also cannot make the `Header` at the highest `z-index` position as modals will be hidden by `Header`.

Z-index priority:
When the APP is connecting an endpoint : Header > Loading 
Modal is opened (Before starting loading): Modal > Header
Modal is opened (After starting loading):    Loading > Modal 

image

impelcrypto: Perhaps I can create a wrapper for modals. Let me try!

impelcrypto: I've fixed in this commit. https://github.com/AstarNetwork/astar-apps/pull/477/commits/0e1794181bb574e01c5f785df8f74e7d9d20b2ea
I think this commit makes more sense.

hoonsubin (approved):

Kahonnohak (approved):
Modal can be always on top of everything, if user wants other action to make such as changing network or accounts then they should click somewhere or press X to close imo:) Great work! I should be more careful for people using 13inch.

PR #479, opened 2022-08-03
feat: added new transfer page

impelcrypto:
**Pull Request Summary**

* feat: added new transfer page
  * path: assets/transfer 

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes
- [x] tested on mobile devices

**Adds**
=Mobile=

=Laptop=

image

**To-dos**
* Delete the legacy code

- [ ] (ex: add user list)


hoonsubin:
Screenshot from 2022-09-01 14-40-24

The text for the side info on the tx page seems very light. Is it a design choice to make the contrast low?

impelcrypto:
@hoonsubin 
> The text for the side info on the tx page seems very light. Is it a design choice to make the contrast low?

Yes, this is according to the design @Kahonnohak

Kahonnohak:
@hoonsubin It is on purpose. This page is for making transactions that are the priority. This place on the right is to give a good experience getting the right information straight away for those who look for it but should be toned down. However, for the Light theme, I do think it might be too thin and I have been thinking to do some colour tests for this Grey 3. It is already used and would have to check how it all looks before making a decision. Let me find out.

hoonsubin:
> @hoonsubin It is on purpose. This page is for making transactions that are the priority. This place on the right is to give a good experience getting the right information straight away for those who look for it but should be toned down. However, for the Light theme, I do think it might be too thin and I have been thinking to do some colour tests for this Grey 3. It is already used and would have to check how it all looks before making a decision. Let me find out.

Yeah, I get the intention, but because of the contrast, I thought the area was deactivated or there was an error at first.

Kahonnohak:
> Yeah, I get the intention, but because of the contrast, I thought the area was deactivated or there was an error at first.

i think it is still ok, people who don't need this area do not need to even notice. People who wander around would hover around. And I will add a little coloured banner to link our youtube that would help the existence if they wanted to notice...

src/components/assets/EvmCbridgeToken.vue
```diff
@@ -124,6 +114,12 @@ export default defineComponent({
       }
     };
 
+    const transferLink = computed<string>(() => {
```
hoonsubin, line 117: This block is repeated a lot. What about making it as a function?

src/components/assets/EvmAssetList.vue
```diff
@@ -249,24 +213,11 @@ export default defineComponent({
...
-        symbol.value = nativeTokenSymbol.value;
-      } else {
-        const c = currency as SelectedToken | Erc20Token;
-        symbol.value = c.symbol;
-      }
-    };
+    const transferLink = computed<string>(() => {
+      const symbol = nativeTokenSymbol.value.toLowerCase();
+      const network = currentNetworkName.value.toLowerCase();
+      return `/assets/transfer?token=${symbol}&network=${network}&mode=local`;
```
hoonsubin, line 219: Another part, since this link has a reference to the network via `network=${network}`, what about moving the network param as part of the base URL. Like `https://portal.astar.network/?rpc=wss%3A%2F%2Fshibuya-rpc.dwellir.com#/assets/transfer/token=${symbol}&mode=local` as an example (like the polkadot js portal).

src/components/assets/transfer/BackToAsset.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <div class="container--back">
+    <router-link to="/assets">
```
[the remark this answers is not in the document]
impelcrypto: Sure but I don't know what `database-like script` looks like. Could you help elaborate on it a little more? Some examples would be helpful.

src/components/assets/EvmAssetList.vue
impelcrypto: Thank you for your suggestion. 

I've checked [documentation](https://router.vuejs.org/guide/essentials/history-mode.html#hash-mode) but I'm afraid Vue-router doesn't work in that way (adding param before '#'). 

Instead, I can set the ':network' [params](https://router.vuejs.org/guide/essentials/dynamic-matching.html) after /assets, so it looks like `/assets/astar/transfer?token=astr&network=astar&mode=local` Looks good to you?

image

src/components/assets/transfer/BackToAsset.vue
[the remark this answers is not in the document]
impelcrypto: Do you mean like [this](https://router.vuejs.org/guide/essentials/named-routes.html)? If so, I've tried but it did't work properly.

https://www.loom.com/share/365a7ec4936c4554a470463008174d39

```
const routes: RouteRecordRaw[] = [
  {
    path: '/assets/:network',
    name: 'Assets',
    component: AssetsPage,
}
```

```
    <router-link :to="{ name: 'Assets', params: { network: 'astar' } }">
      <span> {{ $t('assets.transferPage.backToAssets') }}</span>
    </router-link>
```

 I've also faced quite a few issues related to Vue-router sometimes, not sure if this is because we are using SSR.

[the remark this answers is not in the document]
impelcrypto: I just came up with an alternative solution, let me try it.

Kahonnohak:
@impelcrypto 
Could you please add this MVP vr of text banner please?

Light Theme
https://gyazo.com/1d3708694a6470309a09d98cbfd74659

Dark Theme
https://gyazo.com/e33e616c1cb9bef225555c0e54a03787

Design can be found in the Figma file in the top left frame.

Screenshot 2022-09-02 at 14 14 58

@hoonsubin 
FYI: I'll ask Roy to do some colour tests in a different ticket.

bobo-k2:
> > Yeah, I get the intention, but because of the contrast, I thought the area was deactivated or there was an error at first.
> 
> i think it is still ok, people who don't need this area do not need to even notice. People who wander around would hover around. And I will add a little coloured banner to link our youtube that would help the existence if they wanted to notice...

I noticed that on mobile view contrast is much bigger. The same applies for the dark theme. Maybe to unify those two views.
image

src/hooks/xcm/useTransferRouter.ts
```diff
@@ -0,0 +1,383 @@
...
+  XcmChain,
+  xcmChains,
+  xcmToken,
+} from 'src/modules/xcm';
+import { generateAssetFromEvmToken, generateNativeAsset } from 'src/modules/xcm/tokens';
+import { useStore } from 'src/store';
+import { Asset } from 'src/v2/models';
+import { computed, ref, watch, watchEffect } from 'vue';
+import { useRoute, useRouter } from 'vue-router';
+import { EvmAssets, XcmAssets } from './../../store/assets/state';
```
sirius651, line 19: tiny point : it's better to use not relative path by changing to `src/store/assets/state`

src/hooks/xcm/useXcmEvm.ts
```diff
@@ -1,3 +1,4 @@
+import { addXcmTxHistories } from './../../modules/xcm/utils/index';
```
sirius651, line 1: same here

src/modules/account/utils/index.ts
```diff
@@ -1,6 +1,14 @@
+import { updateAccountHistories } from './../../../config/localStorage';
```
sirius651, line 1: same here

Kahonnohak:
> > > Yeah, I get the intention, but because of the contrast, I thought the area was deactivated or there was an error at first.
> 
> > 
> 
> > i think it is still ok, people who don't need this area do not need to even notice. People who wander around would hover around. And I will add a little coloured banner to link our youtube that would help the existence if they wanted to notice...
> 
> 
> 
> I noticed that on mobile view contrast is much bigger. The same applies for the dark theme. Maybe to unify those two views.
> 
> image
> 
> 

Yes it's how it should be. On mobile, the transfer form is a solo and FAQ sections don't bother the main purpose.

src/components/assets/EvmAssetList.vue
hoonsubin: Hmm, it seems like the string parser does not work well with vue router according to https://github.com/vuejs/vue-router/issues/1259 so the nested query won't work.
Adding a custom parser is too much work too. I like your idea, but instead of putting the network param after the assets, how about moving it to the root (before assets)?

src/components/assets/EvmCbridgeToken.vue
impelcrypto: I've fixed in [this](https://github.com/AstarNetwork/astar-apps/pull/479/commits/09f01d11f8b9a7892d85517e1d40532136f4430f#diff-003c474de6953460a9c7b1afc9a133e05421b357d4bb201c78d9bb2f7f21f30aR16) commit.

src/components/assets/transfer/BackToAsset.vue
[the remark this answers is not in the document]
impelcrypto: I've fixed in [this](https://github.com/AstarNetwork/astar-apps/pull/479/commits/5dcf8e15688d8df1c3f083530ca70a8efed9b24c) commit.

Path:
https://github.com/AstarNetwork/astar-apps/blob/09f01d11f8b9a7892d85517e1d40532136f4430f/src/router/routes.ts#L20

src/components/assets/EvmAssetList.vue
impelcrypto: I've updated the router logic in [this](https://github.com/AstarNetwork/astar-apps/pull/479/commits/09f01d11f8b9a7892d85517e1d40532136f4430f) commit.
So now, the app network can be changed by typing manually like this:

https://astar-apps--pr479-feat-transfer-page-ka2ln004.web.app/#/astar/assets
https://astar-apps--pr479-feat-transfer-page-ka2ln004.web.app/#/shiden/assets
https://astar-apps--pr479-feat-transfer-page-ka2ln004.web.app/#/shibuya-testnet/assets

Kahonnohak: This is cool!

src/hooks/xcm/useTransferRouter.ts
impelcrypto: I've changed almost all the importing styles to the absolute path instead of the relative path in [this](https://github.com/AstarNetwork/astar-apps/pull/479/commits/d356fb33de77d5ce2b34fcbc7406834544f22dfb) commit. But could you help define the rule for importing format? I think we should all be on the same page from now on. @bobo-k2 @hoonsubin @niklabh

src/hooks/xcm/useXcmEvm.ts
impelcrypto: Fixed.

src/modules/account/utils/index.ts
impelcrypto: Fixed.

hoonsubin (approved):
LGTM

amazing work!

bobo-k2 (approved):
LGTM

Kahonnohak (approved):
Great work!

sirius651 (approved):

niklabh (approved):

PR #511, opened 2022-09-14
Wallet services to return transaction hash

bobo-k2:
**Pull Request Summary**

Wallet services to return transaction hash


impelcrypto (approved):
Thank you!

PR #521, opened 2022-09-26
feat: supported XCMP integration with Interlay and Kintsugi

impelcrypto:
**Pull Request Summary**

* feat: supported XCMP transfer between Astar and Interlay
  * feat: added `INTR` amd `IBTC`
* feat: supported XCMP transfer between Shide and Kintsugi
  * feat: added `KINT`, and `KBTC`
* fix: display smaller token amount for wrapped BTC 

**Progresses**
- [x] listing tokens on the native and EVM assets panel
- [x] depositing tokens (to Native/EVM network)
- [x] withdrawing tokens (from Native/EVM network)  
  
**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**

image

image

**Changes**
* fix: display smaller token amount for wrapped BTCs ([code](https://github.com/AstarNetwork/astar-apps/pull/521/files#diff-1081888f2d78afd6f064737cb0b785e8ad4c26ca472179d8b973728533745c33R32))
  * display the account balance without truncating number


src/v2/models/XcmModels.ts
```diff
@@ -23,6 +25,8 @@ export enum parachainIds {
   MOONBEAM = 2004,
   STATEMINE = 1000,
   STATEMINT = 1000,
+  KINTSUGI = 2092,
```
bobo-k2, line 28: Does it make sense to move parachain id to `src/modules/xcm/index.ts `where all parachains are defined?

impelcrypto: Yes, I wanted too, but now I'd like to avoid defining variables under `src/modules/xcm/index.ts` because importing variables from this file cause an error in unit tests.

bobo-k2 (approved):
LGTM

PR #526, opened 2022-09-29
Oak integration

niklabh:
**Pull Request Summary**

**Check list**
- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)


Kahonnohak:
Lovely!

I noticed a few things, Please see below.

<Function>
- Dashboard status doesn't seem to reflect correctly.
- - I've changed my status to "I'd rather manage..." but the dashboard still says "Claim and restake"

- The top-up warning should only appear when you are "Automation" and "the deposit has runout"
- Modal doesn't have a close button. 
- - Please use Astar UI (or if it is easier to add the close button then that is also ok since we change the UI)
- - Also please make the modal close whenever you click other areas. (see current modals)

<UI>
- The "new" tag
- - please add 4px (so it will be 8px) paddings each **side**.
- - font-weight: 700;
- Please align these two
- - Inside of the dashboard, text and the button.
- - The warning message and top-up button.
Screenshot 2022-09-30 at 12 18 17

PR #533, opened 2022-10-11
feat: added `stake` logic

impelcrypto:
**Pull Request Summary**

* feat: added `stake` logic

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)


src/components/dapp-staking/stake-manage/StakeForm.vue
```diff
@@ -64,22 +64,21 @@
...
-        v-if="errMsg && currentAccount"
-        class="row--box-error"
-        :class="isRequiredCheck && 'box--margin-adjuster'"
-      >
+      <div v-if="errMsg && currentAccount" class="row--box-error">
         <span class="color--white"> {{ $t(errMsg) }}</span>
       </div>
       <div class="wrapper__row--button" :class="!errMsg && 'btn-margin-adjuster'">
-        <astar-button class="btn-size--confirm" :disabled="errMsg || !Number(amount)">
+        <astar-button
```
hoonsubin, line 77: I see inconsistent component name format (`kebab-case` and `PascalCase`).

Can we unify them into one and add format checks?

src/components/dapp-staking/stake-manage/SelectFunds.vue
```diff
@@ -14,12 +14,12 @@
...
           <span class="text--item-name">{{ list.name }}</span>
           <div class="column--amount">
-            <span>{{
-              $t('amountSymbol', {
-                amount: $n(truncate(ethers.utils.formatEther(list.balance))),
-                symbol: nativeTokenSymbol,
-              })
-            }}</span>
+            <span>
+              <TokenBalance
```
hoonsubin, line 18: Screenshot from 2022-10-11 11-24-50

Please correct me if I'm wrong, but why can I select different tokens to stake with even though the chain can only support native tokens? Plus, these tokens are ERC20 while I'm logged in with my native account,

The token list also says `Select Chain`

hoonsubin (changes requested):

src/v2/services/implementations/DappStakingService.ts
```diff
@@ -49,12 +49,39 @@ export class DappStakingService implements IDappStakingService {
...
+
+    const stakeCall = await this.dappStakingRepository.getNominationTransferCall({
+      amount,
+      fromContractId,
+      targetContractId,
+    });
+    await this.wallet.signAndSend(
+      stakeCall,
+      address,
+      `You successfully staked to ${targetContractId} from ${fromContractId}`
```
bobo-k2, line 75: Maybe to put this to translation file. I also always forget to do that.

src/v2/services/implementations/DappStakingService.ts
```diff
@@ -49,12 +49,39 @@ export class DappStakingService implements IDappStakingService {
     return new TvlModel(tvl, tvlDefaultUnit, tvlUsd);
   }
 
+  public async nominationTransfer({
```
bobo-k2, line 52: Maybe to write test for this method. Although, there is no much logic to test here you can check if repo method is called with proper parameters, also you can test the case when not parameters are provided

bobo-k2 (changes requested):

src/components/dapp-staking/stake-manage/SelectFunds.vue
Kahonnohak: It's the nomination transfer list + transferable. So the title should be just "Select" or "Use funds from". Thanks for the heads up, actually that makes me think we probably need more information about nomination transfer.

impelcrypto: > Please correct me if I'm wrong, but why can I select different tokens to stake with even though the chain can only support native tokens?

Are you talking about 'JPYC'? The 'JPYC' is the name of dApp that listed on dApp staking, not ERC20 token. 

image


> The token list also says Select Chain

Thank you, it should be 'Select'. I've fixed in [this](https://github.com/AstarNetwork/astar-apps/pull/533/commits/0da79fa712331c0fe3da9ce424bb0bf33e15d3b2) commit
image

src/v2/services/implementations/DappStakingService.ts
impelcrypto: @bobo-k2  True, but I'm wondering how we add `useI18n()` to Vuex files. Do you have any idea?

src/components/dapp-staking/stake-manage/StakeForm.vue
impelcrypto: Ok, Let's use kebab-case from now on.
CC: @bobo-k2 @sirius651 

> Can we unify them into one and add format checks?
@hoonsubin I'll add this on backlog, I will create another PR for this later.

src/v2/services/implementations/DappStakingService.ts
impelcrypto: Otherwise, we can create an argument like `successMsg` for this function but I'd like to avoid this approach if we have another way to fix this.

bobo-k2: Yes, you are right. Now I remember I have the same issue. Let's leave as is for now

src/v2/services/implementations/DappStakingService.ts
impelcrypto: Sure, how about like [this?](https://github.com/AstarNetwork/astar-apps/pull/533/commits/1cf3057c67ef6bf819f4f6755166c0bfaaec9cea)

bobo-k2 (approved):
LGTM

src/components/dapp-staking/stake-manage/ModalSelectFunds.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <AstarModal
+  <astar-modal
```
sirius651, line 2: if not using astar-ui library, please not using like astar-[componentName] . the format already used in ui library as default, it might be conflicts. if you want to implement the common component, feel free to contributing in ui library instead. @impelcrypto @bobo-k2

impelcrypto: @sirius651 Ok. As for AstarModal, this is a wrapper component for astar-simple-modal and it cannot be defined in Astar-UI because I want to use `Teleport`. [Ref](https://github.com/AstarNetwork/astar-apps/blob/main/src/components/common/AstarModal.vue) I'll change this component name.

impelcrypto: @sirius651 I've fixed in [this](https://github.com/AstarNetwork/astar-apps/pull/533/commits/939efb7e595d8c1cc2163a1bdb1f2ff56078a92e) commit.

sirius651 (approved):

src/components/dapp-staking/stake-manage/SelectFunds.vue
hoonsubin: okay, I realized that it was part of the nomination transfer. I think the label was confusing for me.

impelcrypto: @Kahonnohak Let me know if you want to change the design or text

PR #538, opened 2022-10-19
fix: selects statemine on the XCM dropdown

impelcrypto:
**Pull Request Summary**

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
- (ex: Add feature A)

**Fixes**

**Changes**
- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)

### DIFF
--- a/src/components/assets/EvmAssetList.vue
+++ b/src/components/assets/EvmAssetList.vue
@@ -71,7 +71,7 @@
             <div class="column column--balance">
               <div class="column__box">
                 <div class="text--accent">
-                  <TokenBalance :balance="bal" :symbol="nativeTokenSymbol" />
+                  <TokenBalance :balance="String(bal)" :symbol="nativeTokenSymbol" />
                 </div>
                 <div class="text--label">
                   <span>{{ $n(balUsd) }} {{ $t('usd') }}</span>
--- a/src/components/assets/NativeAssetList.vue
+++ b/src/components/assets/NativeAssetList.vue
@@ -25,7 +25,7 @@
             <div class="column--balance">
               <div class="column__box">
                 <div v-if="!isSkeleton" class="text--accent">
-                  <TokenBalance :balance="bal" :symbol="nativeTokenSymbol" />
+                  <TokenBalance :balance="String(bal)" :symbol="nativeTokenSymbol" />
                 </div>
                 <div v-else class="skeleton--right">
                   <q-skeleton animation="fade" class="skeleton--md" />
@@ -81,7 +81,7 @@
             <div class="column--balance">
               <div v-if="!isSkeleton" class="column__box">
                 <span class="text--value">
-                  <TokenBalance :balance="numEvmDeposit" :symbol="nativeTokenSymbol" />
+                  <TokenBalance :balance="String(numEvmDeposit)" :symbol="nativeTokenSymbol" />
                 </span>
               </div>
               <div v-else class="column__box">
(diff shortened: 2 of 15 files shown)


src/components/assets/NativeAssetList.vue
```diff
@@ -25,7 +25,7 @@
             <div class="column--balance">
               <div class="column__box">
                 <div v-if="!isSkeleton" class="text--accent">
-                  <TokenBalance :balance="bal" :symbol="nativeTokenSymbol" />
+                  <TokenBalance :balance="String(bal)" :symbol="nativeTokenSymbol" />
```
hoonsubin, line 28: why is this value nullable?

src/components/assets/NativeAssetList.vue
```diff
@@ -81,7 +81,7 @@
             <div class="column--balance">
               <div v-if="!isSkeleton" class="column__box">
                 <span class="text--value">
-                  <TokenBalance :balance="numEvmDeposit" :symbol="nativeTokenSymbol" />
+                  <TokenBalance :balance="String(numEvmDeposit)" :symbol="nativeTokenSymbol" />
```
hoonsubin, line 84: I see that the `numEvmDeposit` type is `number`. Is this big enough to display the token amount? I maybe overthinking this (in which case, you can safely ignore my comment), but I feel like casting to a string isn't the best way.
What do you think about using `BigNumber` type for all values?

impelcrypto: `numEvmDeposit` is already formatted to ASTR from wei, so we don't need to consider BigNumber (I use BigNumber whenever the app has to calculate with 'wei' unit only).

impelcrypto: There is no case for passing the amount with wei unit to `TokenBalance` component, so I don't think we need to cast the amount to BigNumber. `TokenBalance` displays the amount of the token and unit like `100 ASTR`.

hoonsubin (approved):

src/components/assets/EvmAssetList.vue
```diff
@@ -71,7 +71,7 @@
             <div class="column column--balance">
               <div class="column__box">
                 <div class="text--accent">
-                  <TokenBalance :balance="bal" :symbol="nativeTokenSymbol" />
+                  <TokenBalance :balance="String(bal)" :symbol="nativeTokenSymbol" />
```
bobo-k2, line 74: Instead having String conversion on each TokenBalance component call, can you move this conversion inside the component. So pass number to the component and inside component do String(balance)

bobo-k2 (changes requested):

src/components/assets/EvmAssetList.vue
impelcrypto: @bobo-k2 
No, Vue will display the warning messages whenever passing the wrong format as props to components. 
Actually, what I want is to set the props type as `type: String || Number` like [this](https://github.com/AstarNetwork/astar-apps/pull/538/files#diff-1081888f2d78afd6f064737cb0b785e8ad4c26ca472179d8b973728533745c33L18) but sadly it is not working.


image

bobo-k2: Can you try with type: [String, Number]

impelcrypto: Thank you, but I still got this error...

image

image

bobo-k2 (approved):

PR #570, opened 2022-11-03
Hotfix - loading indicator not visible when modal opens up

sirius651:
**Pull Request Summary**

I found that there's z-index problem between modal and loading indicator. we need resolve it later.
but for now, while transaction on the modal, loading indicator is not visible. because z-index of modal  is more higher than loading indicator one. so to visible loading indicator, opened hot fix for this.

스크린샷 2022-11-03 오후 5 18 46


bobo-k2 (approved):
LGTM

bobo-k2 (approved):

Kahonnohak (approved):
Great Thanks for the fix!

PR #600, opened 2022-11-22
Lightclient UI

gluneau:


src/config/api/polkadot/connectApi.ts
```diff
@@ -152,7 +157,7 @@ export async function connectApi(
     });
   } catch (e) {
     console.error(e);
-    fallbackConnection({ networkIdx, endpoint });
+    //fallbackConnection({ networkIdx, endpoint });
```
bobo-k2, line 160: I think we should fallback to other endpoint in case of error

PR #601, opened 2022-11-23
Developer unique name check

bobo-k2:
**Pull Request Summary**

Fix for a bug in registration page when user was able to give the same name to multiple developers. Since name is the key for search when updating developers, this caused issue with overriding data of other developer and some other problems.

I believe this caused yesterday's issue with XY finance registration.

image

**Check list**
- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


Kahonnohak:
Thank you! It looks great but again I cannot review the code nor on Staging URL so leave it with the other two!

sirius651 (approved):

impelcrypto (approved):

PR #640, opened 2022-12-23
Initial splash screen version

bobo-k2:
**Pull Request Summary**

Added splash screen while waiting for the portal to connect to a node.

**Check list**

- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Release notes:**

- Splash screen added


impelcrypto:
Initial splash screen works great!
I found this error after I transferred tokens.
 
image

bobo-k2:
> Initial splash screen works great! I found this error after I transferred tokens.
This has been fixed here https://github.com/AstarNetwork/astar-apps/pull/643 I will update my branch

impelcrypto:
@bobo-k2 
It doesn't look like transferring is working.
https://www.loom.com/share/5767517eba864371964c8359acb8cf2a

impelcrypto (approved):

sirius651 (approved):

Kahonnohak:
Thanks @bobo-k2 !

This case, the text is white since we have BG colour. 

But my questions are the same
- Loading doesn't take 30s when I select "light clients" and after connecting and checking the Network I seem to be connected to On Finality. So what should I look at? Is this because we are still not connecting to light clients and no need to worry about it? Or it should be connecting to the light clients?

- What is the real estimation time of connecting to light clients? Does it really take 30s?

bobo-k2:
> Hi @bobo-k2 @gluneau Can we merge this branch to `feat/light-client` ? We have to resolve a lot of conflict files on that branch.

It is fine with me. @Kahonnohak had comment to not show data loading animation if we have the splash scree. This means that we need to hide splash after data is loaded. Since data loading is performed by multiple methods on multiple pages (assets, dapp staking, dashboard) we need to do some refactoring. I suggest we think about this and do refactoring outside this PR.

Kahonnohak:
> 

Yes but if we are to launch on Shibuya now I think it is just fine to let people try. But yes if you could look at it in the future that would make it definitely better!

impelcrypto:
Ok, then let me merge this PR first :D

PR #653, opened 2023-01-05
feat: integrate Light client into the portal 

impelcrypto:
**Pull Request Summary**

* feat: integrate Light client into the portal
  * Astar or Shiden: enabled light client endpoint option if the app is opened on Staging URL or Localhost
  * Shibuya: enabled in any URLs

**Check list**

- [ ] contains breaking changes
- [ ] adds new feature
- [ ] modifies existing feature (bug fix or improvements)
- [ ] relies on other tasks
- [ ] documentation changes
- [ ] tested on mobile devices

**Release notes:**

* feat: integrate Light client into the portal

**This pull request makes the following changes:**

**Adds**

- (ex: Add feature A)

**Fixes**

- (ex: Fix validation function)

**Changes**

- (ex: Change document B)

**To-dos**

> \*Feel free to remove this section if it's not applicable

- [ ] (ex: add user list)

PR #656, opened 2023-01-09
Migrate business logic in dapp staking v2 to new architecture

sirius651:
**Pull Request Summary**

> Refactor focusing on these hooks
- useClaimAll
- useStakerInfo
- useUnbonding

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Release notes:**

- refactor: use the new architecture for sending dApp staking transactions

### DIFF
--- a/src/v2/services/IWalletService.ts
+++ b/src/v2/services/IWalletService.ts
@@ -1,4 +1,7 @@
 import { SubmittableExtrinsic } from '@polkadot/api/types';
+import { ISubmittableResult } from '@polkadot/types/types';
+import { SubstrateAccount } from 'src/store/general/state';
+import { HistoryTxType } from 'src/modules/account';
 
 export enum WalletType {
   Metamask = 'Metamask',
@@ -19,4 +22,27 @@ export interface IWalletService {
     successMessage?: string,
     transactionTip?: number
   ): Promise<string | null>;
+
+  signAndSendWithCustomSignature({
+    transaction,
+    senderAddress,
+    substrateAccounts,
+    isCustomSignature,
+    txResHandler,
+    handleCustomExtrinsic,
+    tip,
+    txType,
+  }: {
+    transaction: SubmittableExtrinsic<'promise', ISubmittableResult>;
+    senderAddress: string;
+    substrateAccounts: SubstrateAccount[];
+    isCustomSignature: boolean;
+    txResHandler: (result: ISubmittableResult) => Promise<boolean>;
+    // from: useCustomSignature.ts
+    handleCustomExtrinsic?: (
+      method: SubmittableExtrinsic<'promise', ISubmittableResult>
+    ) => Promise<void>;
+    tip?: string;
+    txType?: HistoryTxType;
+  }): Promise<boolean>;
 }
--- a/src/v2/services/IWalletService.ts
+++ b/src/v2/services/IWalletService.ts
@@ -1,7 +1,4 @@
 import { SubmittableExtrinsic } from '@polkadot/api/types';
-import { ISubmittableResult } from '@polkadot/types/types';
-import { SubstrateAccount } from 'src/store/general/state';
-import { HistoryTxType } from 'src/modules/account';
 
 export enum WalletType {
   Metamask = 'Metamask',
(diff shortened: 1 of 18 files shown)


src/v2/services/IWalletService.ts
```diff
@@ -19,4 +22,27 @@ export interface IWalletService {
     successMessage?: string,
     transactionTip?: number
   ): Promise<string | null>;
+
+  signAndSendWithCustomSignature({
```
impelcrypto, line 26: [signAndSend](https://github.com/AstarNetwork/astar-apps/blob/2f553afa191c401a50cac9dd0bad9a4db5c92be2/src/v2/services/implementations/PolkadotWalletService.ts#L36) method from PolkadotWalletService should work with the lockdrop account. You can take a look at [this](https://github.com/AstarNetwork/astar-apps/pull/653) branch (sending tokens from lockdrop account).

impelcrypto: CC @bobo-k2

impelcrypto: @sirius651 

> [signAndSend](https://github.com/AstarNetwork/astar-apps/blob/2f553afa191c401a50cac9dd0bad9a4db5c92be2/src/v2/services/implementations/PolkadotWalletService.ts#L36) method from PolkadotWalletService

Ah sorry, it should be signAndSend method from [MetamaskWalletService](https://github.com/AstarNetwork/astar-apps/blob/2f553afa191c401a50cac9dd0bad9a4db5c92be2/src/v2/services/implementations/MetamaskWalletService.ts#L35)

Anyway, I think you can use it like [this](https://github.com/AstarNetwork/astar-apps/blob/feat/light-client/src/v2/services/implementations/TokenTransferService.ts#L26)

src/v2/services/implementations/PolkadotWalletService.ts
```diff
@@ -90,6 +92,96 @@ export class PolkadotWalletService extends WalletService implements IWalletServi
     return result;
   }
 
+  // from src/hooks/helper/wallet.ts
+  // @TODO: Need to refactor
+  public async signAndSendWithCustomSignature({
```
bobo-k2, line 97: There is already a very similar signAndSend method in the service. Is it possible to reuse this method here, or extend singAndSend with isCustomSignature?

bobo-k2 (changes requested):

src/v2/services/IWalletService.ts
sirius651: does that mean we can remove useCustomSignature hook further?

impelcrypto: Yes, I think we can remove that hook eventually

src/v2/services/implementations/PolkadotWalletService.ts
sirius651: Yes, that makes sense. as overall, I'd like to ask that do we need to keep [useClaim](https://github.com/AstarNetwork/astar-apps/blob/e3103d8d58e20ac9b29e6e6dd09f3d4ba803e20b/src/hooks/useClaimAll.ts#L27) hook since it has vuex store or other hooks depends on?  otherwise, can we retrieve these all information from v2 architecture whether it already implemented or not?

bobo-k2: In my opinions hooks are fine if they use v2 services

impelcrypto:
Hi @sirius651 
I've created new PR with reusing the v2 signAndSend function [here](https://github.com/AstarNetwork/astar-apps/pull/662). I think this PR make senses and you work in that PR. You can see how I changed useClaimAll function so that you can apply the same methods for useStakerInfo and useUnbonding. You may add some logic in DappStakingRepository.ts as well.

sirius651:
> Hi @sirius651
> I've created new PR with reusing the v2 `signAndSend` method [here](https://github.com/AstarNetwork/astar-apps/pull/662). I think this PR makes sense so how about working in that PR? You can see how I changed `useClaimAll` hook so that you can apply the same way to `useStakerInfo` and `useUnbonding`. You may add some logic in `DappStakingRepository.ts` as well.

Oh, I also same worked in this [commit](https://github.com/AstarNetwork/astar-apps/pull/656/commits/3ea783dc09fbcdce8ee3b192de939a6f28c15e49) . maybe I get used to it now, will move on that PR. following your code is more makes sense

src/v2/services/implementations/ExtrinsicCallService.ts
```diff
@@ -0,0 +1,80 @@
...
+import { ISubmittableResult } from '@polkadot/types/types';
+import { IExtrinsicCallService } from '../IExtrinsicCallService';
+import { IExtrinsicCallRepository } from 'src/v2/repositories/IExtrinsicCallRepository';
+import { BusyMessage, ExtrinsicStatusMessage, IEventAggregator } from 'src/v2/messaging';
+import { WalletService } from './WalletService';
+import { Guard } from 'src/v2/common';
+import { EcdsaAccount } from 'src/store/general/state';
+
+@injectable()
+export class ExtrinsicCallService extends WalletService implements IExtrinsicCallService {
```
sirius651, line 13: @impelcrypto before moving on that PR your made, just wanna check. do you think is it good for making extrinsicCallService like this?

impelcrypto: @sirius651 
I think we don't need to have this `extrinsicCallService` file so that we don't need to make (duplicate) another `send` method 🙂 CC @bobo-k2

PR #677, opened 2023-01-27
Hide notification automatically

sirius651:
**Pull Request Summary**

- Hide notification toast automatically 
- set txHash for success notification

스크린샷 2023-01-27 오전 10 08 41

**Check list**

- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Release notes:**
- feat: hide notification automatically

### DIFF
--- a/src/store/general/actions.ts
+++ b/src/store/general/actions.ts
@@ -13,7 +13,7 @@ const actions: ActionTree<State, StateInterface> = {
 
     setTimeout(() => {
       commit('removeAlertMsg', 0);
-    }, 4000);
+    }, 8000);
   },
   removeAlertMsg({ commit }, { index }) {
     commit('removeAlertMsg', index);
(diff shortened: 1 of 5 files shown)


Kahonnohak (approved):
LGTM!

src/store/general/actions.ts
```diff
@@ -10,6 +10,10 @@ const actions: ActionTree<State, StateInterface> = {
       txHash,
     };
     commit('pushAlertMsg', alert);
+
+    setTimeout(() => {
+      commit('removeAlertMsg', 0);
+    }, 4000);
```
bobo-k2, line 16: What do you think about increasing timeout to 8 or 10 seconds, since there is a option to manually close toast? I think we should  give users more time to eg. copy transaction hash or similar.

sirius651: yes that makes sense. what do you think especially @Kahonnohak  ?

impelcrypto (approved):

src/store/general/actions.ts
Kahonnohak: Lets give it a go then! 8sec? I'm sorry I missed this comment

PR #707, opened 2023-03-08
Improve “claim and restake”

sirius651:
**Pull Request Summary**

Improve “claim and restake”

- Add message on the warning notification to clarify when transferable balance doesn’t have enough balance to make claims. Invalid balance to make a claim. Please add ASTR tokens in to the account.
- When users select “Max” leave `10ASTR` in the account so it will keep the balance for longer period.
    - Add message before the confirm button on Staking page `Account must hold greater than 10ASTR in transferrable when you stake.`

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Release notes:**

fix: claim and restake issues

스크린샷 2023-03-08 오후 12 19 55
스크린샷 2023-03-08 오후 12 19 25

### DIFF
--- a/src/components/dapp-staking/stake-manage/StakeForm.vue
+++ b/src/components/dapp-staking/stake-manage/StakeForm.vue
@@ -70,21 +70,14 @@
         :set-selected-gas="setSelectedTip"
       />
 
-      <div v-if="errMsg && currentAccount" class="row--box-error">
-        <div class="column--title">
-          <span class="text--dot">・</span>
-          <span class="color--white"> {{ $t(errMsg) }}</span>
-        </div>
-      </div>
       <div class="row--box-warning">
         <div class="column--title">
           <span class="text--dot">・</span>
           <span class="color--white"> {{ $t(warningMsg) }}</span>
         </div>
-        <!-- <div v-if="errMsg && currentAccount" class="column--title">
-          <span class="text--dot">・</span>
-          <span class="color--white"> {{ $t(errMsg) }}</span>
-        </div> -->
+      </div>
+      <div v-if="errMsg && currentAccount" class="row--box-error">
+        <span class="color--white"> {{ $t(errMsg) }}</span>
       </div>
       <div class="wrapper__row--button" :class="!errMsg && 'btn-margin-adjuster'">
         <astar-button
--- a/src/components/dapp-staking/stake-manage/StakeForm.vue
+++ b/src/components/dapp-staking/stake-manage/StakeForm.vue
@@ -70,12 +70,19 @@
         :set-selected-gas="setSelectedTip"
       />
 
-      <div v-if="errMsg && currentAccount" class="row--box-error">
+      <div class="row--box-warning">
+        <div class="column--title">
+          <span class="text--dot">・</span>
+          <span class="color--white"> {{ $t(warningMsg) }}</span>
+        </div>
+        <div v-if="errMsg && currentAccount" class="column--title">
+          <span class="text--dot">・</span>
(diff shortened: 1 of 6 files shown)


Kahonnohak:
Thanks @sirius651 

1) Please display the yellow warning as default and at all times on Staking page. At the moment I only see it when press Max button. But Max button limits the amount by itself so what we really need here to let users (especially those who don't use Max button) to know about the rule.

2) The styling is different, please check the asset page (e.g. XCM) if not in Astar UI

3) Please disable the button when users do not meet the criteria. (balance greater than 10ASTR, and the total stake is greater than 500.)
e.g. balance is 1382, and my input is 1380. 
- It should have the button disabled.
- It should have the yellow warning

And for the error message, how can I reproduce or force the situation if there is a way? I could not make my environment for this. Or can @impelcrypto take a look?

Thanks!!

impelcrypto:
> And for the error message, how can I reproduce or force the situation if there is a way? I could not make my environment for this. 

@sirius651 Let's estimate the fee before signing the claim transaction.

1. Get the estimated fee by using [.paymentInfo](https://wiki.polkadot.network/docs/learn-transaction-fees#calcuating-fees-with-polkadot-js) [here](https://github.com/AstarNetwork/astar-apps/blob/main/src/hooks/useClaimAll.ts#L105)

2. Check if `userBalance > (partialFee * 1.5)`. 
  * The reason for * 1.5: actual used fee might more than the estimated fee. Also we have to consider the tips as well.

3. Display the error toast if the result of No.2 is minus. 
  * you can use ` store.dispatch('general/showAlertMsg' ...` 
  
Ref:
image


image

https://shiden.subscan.io/extrinsic/3543829-2

sirius651:
> @sirius651 Let's estimate the fee before signing the claim transactions.

Thanks Fixed in [here](https://github.com/AstarNetwork/astar-apps/pull/707/commits/3c131ea8e44be7ba0110952d1449b1946aef39d2)

Kahonnohak:
Thanks @sirius651 ! On top of what Roy suggest,

So the first warning `Account must hold greater than 10ASTR in transferrable when you stake` is good!

When pressing Max button and the amount was not enough then `
The amount of token to be staking must be greater than 500 ASTR` needs to be red warning not in the same yellow.

Thanks, I'm still trying to reproduce the situation I can get the error with not enough gas, will try local node and come back here tomorrow.

sirius651:
> Thanks @sirius651 ! On top of what Roy suggest,
> 
> So the first warning `Account must hold greater than 10ASTR in transferrable when you stake` is good!
> 
> When pressing Max button and the amount was not enough then ` The amount of token to be staking must be greater than 500 ASTR` needs to be red warning not in the same yellow.


스크린샷 2023-03-09 오후 11 32 35

Okay does that mean like this?

impelcrypto (changes requested):

Kahonnohak:
> > Thanks @sirius651 ! On top of what Roy suggest,
> > So the first warning `Account must hold greater than 10ASTR in transferrable when you stake` is good!
> > When pressing Max button and the amount was not enough then ` The amount of token to be staking must be greater than 500 ASTR` needs to be red warning not in the same yellow.
> 
> 스크린샷 2023-03-09 오후 11 32 35
> 
> Okay does that mean like this?

Yes! Thanks! And just bullet points are different from the asset page that Roy pointed out. This really is good if we have it in Astar UI at somepoint

impelcrypto (approved):

Kahonnohak (approved):
Great work!

PR #740, opened 2023-04-12
Don't set Ethereum Extension value to currentAddress field

bobo-k2:
**Pull Request Summary**

This PR is to fix issue with using dApps staking with lockdrop accounts.

The root cause of the issue is that `currentAddress` is incorrectly set. `currentAddress` should have the same value as `ss58` field (edited) 
image

I disabled setting 'Ethereum Extension' value to the `currentAddress` field. I am not sure why we are storing such invalid value to Vuex. Also, not sure if this change breaks something else, so serious testing is required.

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Release notes:**

- Fixed dApps staking not working with Lockdrop accounts issue.

### DIFF
--- a/src/hooks/useClaimAll.ts
+++ b/src/hooks/useClaimAll.ts
@@ -15,7 +15,6 @@ import { IDappStakingService } from 'src/v2/services';
 import { Symbols } from 'src/v2/symbols';
 import { ethers } from 'ethers';
 import { computed, ref, watchEffect } from 'vue';
-import { useNetworkInfo } from 'src/hooks';
 import { useI18n } from 'vue-i18n';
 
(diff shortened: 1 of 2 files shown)


impelcrypto:
Let's take a look where [ETHEREUM_EXTENSION](https://github.com/AstarNetwork/astar-apps/blob/main/src/hooks/useAccount.ts#L7) is used (like [here)](https://github.com/AstarNetwork/astar-apps/blob/main/src/hooks/useAccount.ts#L43)

impelcrypto (approved):

sirius651 (approved):

PR #778, opened 2023-05-09
chopsticks e2e poc

gluneau:
**Pull Request Summary**

> Describe what changes this pull request makes to the repository

**Check list**

- [ ] contains breaking changes
- [x] adds new feature
- [ ] modifies existing feature (bug fix or improvements)
- [ ] relies on other tasks
- [ ] documentation changes
- [ ] tested on mobile devices

**Release notes:**

- (ex: feat: updated dApp staking to V2)
- (ex: fix: modified styling bugs on the Assets page)

**This pull request makes the following changes:**

**Adds**

- (ex: Add feature A)

**Fixes**

- (ex: Fix validation function)

**Changes**

- (ex: Change document B)

**To-dos**

> \*Feel free to remove this section if it's not applicable

- [ ] (ex: add user list)

PR #789, opened 2023-05-15
e2e testing ::AS002/AS004/XCM007

sirius651:
**Pull Request Summary**

based on this [PR](https://github.com/AstarNetwork/astar-apps/pull/778)
E2E testing with chopstick : AS002/AS004/XCM007

regarding AS002, there's an issue where the balance between from and to of metamask is different, so this should be solved in another PR.

스크린샷 2023-05-29 오전 11 06 15

**Check list**

- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


sirius651:
Hi, @bobo-k2 do you have any idea why it can't detect `getWindow metamask` [here](https://github.com/AstarNetwork/astar-apps/blob/testing/e2e-will/tests/common.ts#L100)? it just crashed as `getWindow metamask not found`

bobo-k2:
> Hi, @bobo-k2 do you have any idea why it can't detect `getWindow metamask` [here](https://github.com/AstarNetwork/astar-apps/blob/testing/e2e-will/tests/common.ts#L100)? it just crashed as `getWindow metamask not found`

Not at first look, I will checkout and try. Did you try to debug?

sirius651:
> Not at first look, I will checkout and try. Did you try to debug?

yes, it's debug mode

sirius651:
> Hi, @bobo-k2 do you have any idea why it can't detect `getWindow metamask` [here](https://github.com/AstarNetwork/astar-apps/blob/testing/e2e-will/tests/common.ts#L100)? it just crashed as `MetaMask window not found`

sirius651:
Even I tried to change manifest version from 3 to 2, it returned same issue.

bobo-k2:
> Even I tried to change manifest version from 3 to 2, it returned same issue.

Window name is `MetaMask Notifications`, this could be the reason for the issue.

image

sirius651:
> > Even I tried to change manifest version from 3 to 2, it returned same issue.
> 
> Window name is `MetaMask Notification`, this could be the reason for the issue.
> 

thanks, I found that what the issue is. the thing is that playwright can't detect right away when extension loaded. so I can resolve it by waiting for loading completely on the DOM in extension

gluneau (approved):
LGTM

tests/zombienet/setup-preconditions-api.js
```diff
@@ -0,0 +1,36 @@
...
+  const api = await ApiPromise.create({ provider });
+
+  await api.isReady;
+
+  // account to submit tx
+  const keyring = new Keyring({ type: 'sr25519' });
+  const sender = keyring.addFromUri('//' + args[0]);
+
+  // create asset
+  console.log('Creating asset with sender: ', sender.address);
```
impelcrypto, line 23: How about using `console.info` instead of `console.log` for those we want to have it (remain in the repo)
CC: @bobo-k2 @gluneau

impelcrypto (changes requested):

tests/zombienet/setup-preconditions-api.js
sirius651: Cleaned up at this [commit](https://github.com/AstarNetwork/astar-apps/pull/789/commits/a277bbec7d544ec3a31ff27fe2f26edd1fafe517)

sirius651:
If we decide to use chopsticks, is it better to remove the existing zombienet-related code? If our testing works with zombienet, I think it's fine to leave it in place. @bobo-k2 @gluneau @impelcrypto

tests/common.ts
```diff
@@ -24,6 +25,35 @@ export const createAccount = async (page: Page, seed: string, name: string): Pro
   await page.getByRole('button', { name: 'Add the account with the supplied seed' }).click();
 };
 
+export const connectToNetwork = async (page: Page): Promise<void> => {
```
bobo-k2, line 28: Correct me if I am wrong, but I recall we have methods below in common, why moving them to test file?

bobo-k2 (changes requested):

tests/common.ts
sirius651: Sorry, I didn't get your point fully. I just changed the location to separate the metamask related code.

bobo-k2: I am sorry. My mistake, somehow I figured out that the code is in àssets.spec.ts`

bobo-k2 (approved):
LGTM

impelcrypto:
> If we decide to use chopsticks, is it better to remove the existing zombienet-related code? 

Yes, I want to remove it if we won't use it anymore. What do you think? @bobo-k2 @gluneau

impelcrypto (approved):

PR #909, opened 2023-08-22
fix: updated withdrawal UI

impelcrypto:
**Pull Request Summary**

* fix: updated withdrawal UI 
* fix: removed IS_APPLIED_RANDOM_ENDPOINT from local storage

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Changes**
* fix: updated withdrawal UI  
  * if the resulting staking amount is lower than minimum staking amount: 
    * display a warning message
    * unstake all of staked balance
    
Screenshot 2023-08-22 at 4 00 12 PM

### DIFF
--- a/tests/common-api.ts
+++ b/tests/common-api.ts
@@ -35,7 +35,7 @@ export const getBalance = async (address: string, assetId?: string): Promise<big
     balance = BigInt(balanceObj.unwrap().balance.toBigInt());
   } else {
     const balanceObj = (await api.query.system.account(address)) as FrameSystemAccountInfo;
-    balance = BigInt(balanceObj.data.free.toBigInt() - balanceObj.data.flags.toBigInt());
+    balance = BigInt(balanceObj.data.free.toBigInt() - balanceObj.data.frozen.toBigInt());
   }
 
   return balance;
--- a/tests/common-api.ts
+++ b/tests/common-api.ts
@@ -1,3 +1,4 @@
+import { GeneralStakerInfo, getDappAddressEnum } from '@astar-network/astar-sdk-core';
 import { ApiPromise, Keyring, WsProvider } from '@polkadot/api';
 import { Option, u32 } from '@polkadot/types';
 import { AccountLedger } from 'src/hooks';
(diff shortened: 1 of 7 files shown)


tests/common-api.ts
```diff
@@ -34,7 +35,7 @@ export const getBalance = async (address: string, assetId?: string): Promise<big
     balance = BigInt(balanceObj.unwrap().balance.toBigInt());
   } else {
     const balanceObj = (await api.query.system.account(address)) as FrameSystemAccountInfo;
-    balance = BigInt(balanceObj.data.free.toBigInt() - balanceObj.data.feeFrozen.toBigInt());
+    balance = BigInt(balanceObj.data.free.toBigInt() - balanceObj.data.flags.toBigInt());
```
bobo-k2, line 38: I don't think the result is correct in this case since `flags` doesn't represent a Balance. See https://github.com/paritytech/substrate/blob/b7c18f1f5c9dc5c4ffa683e230303f084efe82cb/frame/balances/src/types.rs#L108

bobo-k2 (changes requested):

tests/common-api.ts
bobo-k2: Not just data structure was changed, also meaning of values have been changed. I believe you should use `frozen` property for this calculation.

Since Astar runtime is updated I will make some changes on the `AccountData` on the Portal today or tomorrow. The fix I made after Shibuya runtime upgrade was just temporary.

impelcrypto: @bobo-k2  CC: @gluneau 
Does this PR makes sense to you?
https://github.com/AstarNetwork/astar-apps/pull/910/files

impelcrypto: also, I've updated to use `frozen` in this branch
https://github.com/AstarNetwork/astar-apps/pull/909/commits/4e6ca4b11b848bc5f1c1e36ebc7f5fb5d01e1914

bobo-k2 (approved):
LGTM

gluneau (approved):
Works great, all tests in dappstaking-transactions.spec.ts passed!

PR #914, opened 2023-08-31
Token allocation chart

bobo-k2:
**Pull Request Summary**

image

**Check list**

- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**

- (ex: Add feature A)

**Fixes**

**Changes**

- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)


Kahonnohak:
Looks good! 

So the main issue here is the Others still including some tokens from circulation supply. Please check

- According to our current Portal, Others should be 37% and Circulation should be 63%, those need to be dynamic. 
- In side of TVL it is nice if we know how much are unbonding but not "must" 
Screenshot 2023-09-01 at 09 09 38
Screenshot 2023-09-01 at 09 12 03

bobo-k2:
> https://astar-apps--pr914-feat-token-chart-g9s5kgen.web.app



> Looks good!
> 
> So the main issue here is the Others still including some tokens from circulation supply. Please check
> 
> * According to our current Portal, Others should be 37% and Circulation should be 63%, those need to be dynamic.
> * In side of TVL it is nice if we know how much are unbonding but not "must"
> 
Yes, I am aware of that, but the issue happens because I am missing data for some token allocations. In reality, we have the following situation
37% - others
44% - tvl
4% - treasury
15% - unknown

I can add one more section to the chart for this unknown allocations, but not sure how to call it

bobo-k2:
Added 4th segment. Also added two decimal point to percentage, otherwise percentage rounding make such mistake that at the end percentages sum up to 101%

image

Kahonnohak:
Right!

1, Wording
In the pie chart
Not Circulation Supply -> Others
Not categorised (Other currently) -> Circulating
TVL-> Staking

The out side of the chart, right bottom
Circulating
Others

I added (s) to Other.

That should be aligned and okay!


2, Could you also adjust padding a bit, I could not see the exact numbers but just a little..

Screenshot 2023-09-01 at 16 05 13

Kahonnohak:
Bobo, sorry, I was not looking at the Light theme.

Could you change the Other colour in the chart from Navy to Grey001 F7F7F8. It might be too light but I could not check the change in the inspect🙇

Kahonnohak (approved):
LGTM!

impelcrypto (approved):

gluneau (approved):
LGTM, it works.

PR #931, opened 2023-09-19
Set identity logic

bobo-k2:
**Pull Request Summary**

Implemented logic for the following
- account unification, together with set identity (those two can't be tested as a single batch atm because local node with accounts palled doesn't have identity pallet. Separate tests worked for me)
- NFT repository, to fetch tokens owned by an user

**Check list**

- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**

- (ex: Add feature A)

**Fixes**

**Changes**

- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)


src/v2/repositories/implementations/BluezNftRepository.ts
```diff
@@ -0,0 +1,28 @@
+// TODO move to Token API
+import axios from 'axios';
+import { injectable } from 'inversify';
+import { NftMetadata } from 'src/v2/models';
+import { INftRepository } from 'src/v2/repositories';
+
+@injectable()
+export class BluezNftRepository implements INftRepository {
+  private readonly baseUrl = 'https://api.bluez.app/api/nft/v3/API_KEY';
```
gluneau, line 9: API Key seems required for this "Set Identity"

src/components/assets/Assets.vue
```diff
@@ -5,6 +5,7 @@
     </div>
     <div class="container--assets">
       <astar-button @click="unify()">Unify accounts</astar-button>
+      <astar-button @click="setIdentity()">Set setIdentity</astar-button>
```
gluneau, line 8: An extra 'set'

gluneau (changes requested):
Few changes, but I get an error while clicking on Unify accounts on all networks.
image

bobo-k2:
> Few changes, but I get an error while clicking on Unify accounts on all networks.
> 
> image
> 
> 

Unfortunately you can't test the code from UI yet, because we don't have all pieces connected yet

src/components/assets/Assets.vue
bobo-k2: Those are temporary buttons so I can test pieces of logic

gluneau (approved):
Will follow changes in account unification branch/PR

PR #947, opened 2023-10-02
fix: let space wrap and height grow for cards in banner

gluneau:
**Pull Request Summary**

The Issue: 
image

The Fix:
image

And when there is room, it's just like the old days:
image

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**

- (ex: Add feature A)

**Fixes**

**Changes**

- (ex: Change document B)

**To-dos**

- [ ] (ex: add user list)

### DIFF
--- a/src/components/dapp-staking/my-staking/components/banners/Algem.vue
+++ b/src/components/dapp-staking/my-staking/components/banners/Algem.vue
@@ -41,7 +41,7 @@ export default defineComponent({
   cursor: pointer;
   background: #d9d9d9;
   background-size: cover;
-  height: 127px;
+  min-height: 127px;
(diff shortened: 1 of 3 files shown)


Kahonnohak:
Thanks @gluneau 

It looks too narrow, it was something like this width. I adjusted using width: 130% (~140% would be fine) something similar would be good (somewhere around 160 ~170px is the minimum width)

Also please align the height. 

Thats from me thanks so much

Screenshot 2023-10-02 at 22 35 11

impelcrypto (approved):
LGTM

Kahonnohak (approved):
Thank you!

PR #1003, opened 2023-11-01
fix: Account Unification errors

impelcrypto:
**Pull Request Summary**

* fix: displaying account icon
* fix: use unified account
  * map to proper SS58 address for EVM dApp staking
  * map to proper H160 address for sending assets from EVM to Native(unified)

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Fixes**
* fix: displaying account icon

=Before=
* Not unified account but `isUnifiedAccount` was true
image

* This is unified account but `AuIcon` component was not imported
Screenshot 2023-11-01 at 3 10 05 PM

=After=
image

**Changes**

* removed `getSS58Address` function from `useAccount` hooks and use `getMappedNativeAddress` from `AccountUnificationService`

### DIFF
--- a/src/v2/repositories/implementations/AccountUnificationRepository.ts
+++ b/src/v2/repositories/implementations/AccountUnificationRepository.ts
@@ -1,3 +1,4 @@
+import { buildEvmAddress, toSS58Address } from '@astar-network/astar-sdk-core';
 import { AccountId32, H160 } from '@polkadot/types/interfaces';
 import { inject, injectable } from 'inversify';
 import { Guard } from 'src/v2/common';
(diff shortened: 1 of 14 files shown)


gluneau:
Well done.

image

gluneau (approved):

Kahonnohak:
Checked if account is not unified, portal doesn't allow transactions to happen
Screenshot 2023-11-01 at 15 22 52

Checked if account is unified
Sent ERC20 to native address and received the token in unified account.

src/v2/repositories/implementations/AccountUnificationRepository.ts
```diff
@@ -29,24 +30,37 @@ export class AccountUnificationRepository implements IAccountUnificationReposito
...
     const evmAddress = api.query.hasOwnProperty('unifiedAccounts')
       ? await api.query.unifiedAccounts.nativeToEvm<H160>(nativeAddress)
-      : '';
+      : buildEvmAddress(nativeAddress);
 
-    return evmAddress.toString();
+    return evmAddress.toString() !== '' ? evmAddress.toString() : buildEvmAddress(nativeAddress);
+  }
+
+  public async handleCheckIsUnifiedAccount(address: string): Promise<boolean> {
```
bobo-k2, line 51: This is a general comment for this PR
I see a lot of logic has been added to fetch and check if account is unified. I believe `useAccountUnification` hook can be used with existing logic implemented there
image

When user connects wallet portal already checks if account is unified and stores info in Vuex, so you can reuse data from there for your logic.

bobo-k2 (changes requested):

src/v2/repositories/implementations/AccountUnificationRepository.ts
impelcrypto: Thanks for checking. `isAccountUnified` only can check the sender account but we also have to check if the destination account has already been unified when sending ERC20 tokens from EVM to SS58. Therefore, `handleCheckIsUnifiedAccount` function will check it.

impelcrypto: https://github.com/AstarNetwork/astar-apps/pull/1003#issuecomment-1789169450

bobo-k2: > Thanks for checking. This is for sending tokens. `isAccountUnified` only can check the sender account but we also have to check if the destination account has already been unified when sending ERC20 tokens from EVM to SS58. Therefore, `handleCheckIsUnifiedAccount` function will check it.

Ok, I understand. For checking receiver is fine. Please use `useAccountUnification` hook for sender where applicable

bobo-k2 (approved):

PR #1032, opened 2023-11-16
Allow Astar Wallet Snap on all networks

gluneau:
**Pull Request Summary**

Allow Astar Wallet Snap on all networks

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/src/modules/snap/index.ts
+++ b/src/modules/snap/index.ts
@@ -8,7 +8,8 @@ import {
   isPolkadotSnapInstalled,
 } from '@astar-network/metamask-astar-adapter/build/utils';
 import type { UnitConfiguration } from '@astar-network/metamask-astar-types';
-import { useNetworkInfo } from '../../hooks/useNetworkInfo';
+import { LOCAL_STORAGE } from 'src/config/localStorage';
+import { getNetworkName } from 'src/config/chainEndpoints';
 
 // Todo: move to Astar.js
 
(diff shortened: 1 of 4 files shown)


impelcrypto:
@gluneau 
>    await enablePolkadotSnap({ networkName: 'shibuya' }, snapId);

Do we want to update [here](https://github.com/AstarNetwork/astar-apps/blob/feat/enable-snap-mainnets/src/modules/snap/index.ts#L59) as well?

src/modules/snap/index.ts
```diff
@@ -87,7 +87,7 @@ export async function initiatePolkdatodSnap(): Promise<SnapInitializationRespons
   try {
     console.info('Attempting to connect to snap...');
     // Todo: update network params
-    const metamaskPolkadotSnap = await enablePolkadotSnap({ networkName: 'shibuya' }, snapId);
+    const metamaskPolkadotSnap = await enablePolkadotSnap({ networkName: 'astar' }, snapId);
```
impelcrypto, line 90: @gluneau am I correct that it doesn't matter passing astar while the portal connects to 'Shiden' or 'Shibuya'? Otherwise, you can use networkNameSubstrate from useNetworkInfo hook. You might want to check the actual value under each networks and make it lower case.

gluneau: I've introduced the `networkNameSubstrate` but it seems that the Store or getters not defined yet.

Let me know.

bobo-k2 (approved):

PR #1072, opened 2023-12-04
Improve polkadot.js errors handling

bobo-k2:
**Pull Request Summary**

This PR is to better handle errors before extrinsic call is initiated. E.g. if caller balance is to low to pay gas spinner on UI will spin forever and error will be only visible in a browser console. 

Basically just added one more try/catch block.

BEFORE
image

AFTER
image

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


Kahonnohak:
Thank you, is there any cases other than balance is too low to pay the gas fee?
If not, maybe better to just say
`1010: Invalid transactions: transferable balance is too low to pay some fees`
?

impelcrypto (approved):

PR #1102, opened 2023-12-28
feat: added EVM supports

impelcrypto:
**Pull Request Summary**

* feat: added EVM supports
  * ref: https://github.com/AstarNetwork/astar-apps/pull/1101 

**Check list**

- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

- [ ] (ex: add user list)


src/staking-v3/logic/services/DappStakingServiceEvm.ts
```diff
@@ -0,0 +1,244 @@
...
+    amount: number,
+    senderAddress: string,
+    successMessage: string
+  ): Promise<void> {
+    const ss58Address = await this.getSS58Address(senderAddress);
+    Guard.ThrowIfUndefined(contractAddress, 'contractAddress');
+    Guard.ThrowIfUndefined(senderAddress, 'senderAddress');
+    Guard.ThrowIfNegative('amount', amount);
+
+    const claimStakerCalls = await this.getClaimStakerAndBonusRewardsCalls(ss58Address);
```
bobo-k2, line 40: This code block is the same as in native service. I suggest to refactor this to avoid code duplication. See `claimLockAndStake` method for example. Basically extract duplicated code block to protected method and put it to native service.
Do similar for other calls if there is repeating code.

src/staking-v3/logic/services/DappStakingServiceEvm.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  // @inheritdoc
+  public async claimUnstakeAndUnlock(
+    contractAddress: string,
+    amount: number,
+    senderAddress: string,
+    successMessage: string
+  ): Promise<void> {
+    const ss58Address = await this.getSS58Address(senderAddress);
```
bobo-k2, line 35: `Guard.ThrowIfUndefined(senderAddress, 'senderAddress');` should go above this line

bobo-k2 (changes requested):

bobo-k2:
Just noticed that ledger in Vuex is empty after staking when EVM account is connected 
image

src/staking-v3/logic/services/DappStakingServiceEvm.ts
impelcrypto: Fixed in [this](https://github.com/AstarNetwork/astar-apps/pull/1102/commits/0494d397fc98f0913afaad8d7781d1525312800e) commit.

src/staking-v3/logic/services/DappStakingServiceEvm.ts
impelcrypto: Sorry but I'm afraid I don't understand your point. 'claimStakerCalls' is required for the batch call, therefore I don't think we can make it more simpler here.

impelcrypto: Are you suggesting me to move the logic into `DappStakingRepository` and then refactor both 'DappsStakingServiceEvm' and 'DappStakingService'?

impelcrypto: Ok, will do that

bobo-k2:
> Just noticed that ledger in Vuex is empty after staking when EVM account is connected

Also staker info is not stored. I think I know what is the problem. Let me fix that.

bobo-k2:
Claim call is failing. Not sure why atm.
image

src/staking-v3/logic/services/DappStakingServiceEvm.ts
impelcrypto: Fixed in [this](https://github.com/AstarNetwork/astar-apps/pull/1102/commits/9ef93188ebd7d804e4b115a211cea97a7cfb29c5) commit.

bobo-k2 (approved):

PR #563, opened 2022-10-28
Add Hana wallet support

tzechuen:
Hi, I'm a developer for Hana wallet, currently available on the Chrome web store. 

We would like to add our wallet to the list of supported wallets for Astar portal. 

Please let me know if you have any questions.

Thanks,
Ting

PR #1105, opened 2023-12-30
feat: add Enkrypt support

kvhnuke:
**Pull Request Summary**

> Adds support for Enkrypt Extension

**Check list**

- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**

-  Enkrypt support


Kahonnohak:
Thank you so much for the PR, would it be possible to create a staging link please? Happy New Year and all the best for 2024!

gluneau:
Check here @kvhnuke for an example on how to create a staging preview for us to review.

https://docs.astar.network/docs/build/builder-guides/integration_toolings/deploy-astar-portal

kvhnuke:
https://astar-apps-bmlxvgpbt-kosala-hemachandras-projects.vercel.app/

gluneau:
> https://astar-apps-bmlxvgpbt-kosala-hemachandras-projects.vercel.app/

image

gluneau (changes requested):
You need different source names for native and evm

gluneau (approved):
It works, thanks for the PR.

impelcrypto (changes requested):
The portal doesn't display the loading animation while waiting for finalizing the transaction. I think it's related to the same issue Hana wallet had before. Please take a look at comments below and resolve the issue on the wallet extension. Please let me know once Enkrypt extension has updated version after resolve the issue.

https://github.com/AstarNetwork/astar-apps/pull/563#issuecomment-1517352791

https://github.com/AstarNetwork/astar-apps/pull/563#issuecomment-1533991693

https://github.com/AstarNetwork/astar-apps/pull/563#issuecomment-1546524311

Demo:
https://www.loom.com/share/f6055ec55cda4a719e03922e52987c14?from_recorder=1&focus_title=1

=Polkadot.js=
The animation should work like this:
image

kvhnuke:
@impelcrypto after carefully reviewing how you handle the loading screen, I have to disagree with this approach.
as a DApp you should not listen to "messages" emitted by window event due to following reasons
1. This is not a documented approach
2. window events are inherently insecure and only reason we are still using it is due to the fact the modern communication streams such as "https://developer.chrome.com/docs/extensions/reference/api/runtime#method-sendMessage" are not implemented across all the browsers. However, when it happens we will move away from window messaging.
3. As a dapp you should never implement functions that are not documented as they can easily change.
4. you might get false positive loading screens, if there is a message with `data.response` and `data.signature` (https://github.com/Hana-Technology/astar-apps/blob/add-hana-wallet/src/v2/services/implementations/PolkadotWalletService.ts#L179)
5. if you look closely at https://github.com/polkadot-js/extension/blob/297b2af14c68574b24bb8fdeda2208c473eccf43/packages/extension/src/page.ts#L10-L22 they only use the window message listener to `handleResponse` back to the dapp. 
6. This listener is never intended to be part of a dapp

I do understand you got lucky with talisman.subwallet etc as they probably just copy pasted polkadotjs communication method, however our messages are more indepth and made to custom to Enkrypt. 

Following are few ways you can implement the fix for loading screen on your end
1. Use promises, this is the way https://polkadot.js.org/apps/ handles loading and the documented way at https://polkadot.js.org/docs/extension

if there is any standard or documented method we missed or improperly implemented please let us know, if not this is not a change we can comply with

impelcrypto (approved):

impelcrypto:
@kvhnuke Thanks for letting me know.

PR #1116, opened 2024-01-09
fix: toast for `You do not have enough tokens to pay the transaction fee`

gluneau:
**Pull Request Summary**

- We now check to see if your account balance is less than 0.05 token.
- If it is, it will display an error saying _You do not have enough tokens to pay the transaction fee_
- And it will not proceed with the transaction.
- Both for EVM and native transactions

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


src/v2/services/implementations/PolkadotWalletService.ts
```diff
@@ -77,6 +77,14 @@ export class PolkadotWalletService extends WalletService implements IWalletServi
           tip = tip ? ethers.utils.parseEther(tip).toString() : '1';
         }
 
+        const useableBalance = await this.assetsRepository.getNativeBalance(senderAddress);
+        const message = 'Minimum balance on the network is 0.01';
```
impelcrypto, line 81: Let's add it [here](https://github.com/AstarNetwork/astar-apps/blob/4718312b815cbb895dc08c19bbd249b397ffd535/src/modules/toast/index.ts#L1)

src/v2/services/implementations/PolkadotWalletService.ts
```diff
@@ -77,6 +77,14 @@ export class PolkadotWalletService extends WalletService implements IWalletServi
...
+        if (Number(useableBalance) < Number(100000000000000000)) {
```
impelcrypto: Perhaps the number is too big to calculate. Let's convert the decimals before compare these values.  Maybe you can do something like this:
```
 // REQUIRED_MINIMUM_BALANCE: 0.05
  const isBalanceEnough =
          Number(ethers.utils.formatEther(useableBalance)) > REQUIRED_MINIMUM_BALANCE;
 if (isBalanceEnough) {
   // Logic goes here
        }
```

Also, let's define the REQUIRED_MINIMUM_BALANCE in somewhere (perhaps at `modules/extrinsic/index.ts` or at `modules/toast/index.ts`).

image

Kahonnohak:
Thank you!

Hmmm,

local transfer I still have 

https://gyazo.com/cc5416a2a4feb2f79e11a34f16a8d515

claim, I did get the toast but I had more than 0.05. 

Screenshot 2024-01-10 at 00 26 30

Can we have the same messages of lacking the balance to pay the fee?

bobo-k2 (changes requested):
So we are checking if user balance is greater than 0.05. I believe the main reason is because we want user to have some tokens in the future to be able to pay transaction fees.
What we are not checking here is how much tokens will remain in the user wallet after transaction (e.g. balance transfer).

src/modules/toast/index.ts
```diff
@@ -4,4 +4,7 @@ export enum AlertMsg {
   ERROR = 'Transaction failed',
   ERROR_DUPLICATED_TX = 'Transaction failed due to duplicate transactions on PolkaSafe; please approve or reject them before creating a new transaction',
   COMPLETED_HASH = 'Completed at transaction hash',
+  MINIMUM_BALANCE = 'You do not have enough tokens to pay the transaction fee',
 }
+
+export const REQUIRED_MINIMUM_BALANCE: number = 0.05;
```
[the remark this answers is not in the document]
gluneau: The number gets too big, @impelcrypto asked for it to be this way.

src/v2/services/implementations/PolkadotWalletService.ts
```diff
@@ -77,6 +77,17 @@ export class PolkadotWalletService extends WalletService implements IWalletServi
           tip = tip ? ethers.utils.parseEther(tip).toString() : '1';
         }
 
+        const useableBalance = await this.assetsRepository.getNativeBalance(senderAddress);
+        const isBalanceNotEnough =
+          Number(ethers.utils.formatEther(useableBalance)) < REQUIRED_MINIMUM_BALANCE;
+        if (isBalanceNotEnough) {
+          this.eventAggregator.publish(
```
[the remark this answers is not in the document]
gluneau: I'm getting only a single toast.  Can you show how you get this?

image

src/modules/toast/index.ts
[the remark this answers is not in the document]
impelcrypto: @bobo-k2 
Please refer to [this](https://github.com/AstarNetwork/astar-apps/pull/1116#discussion_r1446141900)

I personally prefer to use ETH instead of WEI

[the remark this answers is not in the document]
bobo-k2: By converting to ETH you are loosing a number precision (in our case balance expressed in wei has 18 decimal places, while balance expressed in ETH/number has 15 to 17 decimal places) and some edge cases will not work. [ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number)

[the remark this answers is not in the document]
impelcrypto: Ok, then we should use WEI for this case. We might want to use `bigint` to compare it with the balance rather than Number. @gluneau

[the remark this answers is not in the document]
gluneau: Let's use `ethers.utils.formatUnits(useableBalance, 18)`

src/v2/services/implementations/AssetsService.ts
```diff
@@ -26,33 +28,51 @@ export class AssetsService implements IAssetsService {
...
   public async transferNativeAsset(param: ParamAssetTransfer): Promise<void> {
-    const transaction = await this.AssetsRepository.getNativeTransferCall(param);
-    const hash = await this.wallet.signAndSend({
-      extrinsic: transaction,
-      senderAddress: param.senderAddress,
-      successMessage: param.successMessage,
-    });
-    param.finalizedCallback(String(hash));
+    const useableBalance = await this.AssetsRepository.getNativeBalance(param.senderAddress);
+    const isBalanceNotEnough =
```
impelcrypto, line 32: Let's change this to `isBalanceEnough`. Having 'Not' in the name makes it less readable.

impelcrypto (changes requested):

src/modules/toast/index.ts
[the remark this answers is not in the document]
impelcrypto: @gluneau Or you can use `ethers.utils.formatEther(wei)` so that you don't need to pass `18`.

src/v2/services/implementations/AssetsService.ts
```diff
@@ -26,33 +28,51 @@ export class AssetsService implements IAssetsService {
...
     const provider = getEvmProvider(this.currentWallet as any);
     const web3 = new Web3(provider as any);
-    const rawTx = await this.AssetsRepository.getEvmTransferData({
-      param,
-      web3,
-    });
 
-    const isErc20 = !!rawTx.data;
+    const balWei = await web3.eth.getBalance(param.senderAddress);
+    const useableBalance = Number(ethers.utils.formatUnits(balWei, 18));
```
impelcrypto, line 55: And here too
https://github.com/AstarNetwork/astar-apps/pull/1116/files#r1448571612

src/v2/services/implementations/AssetsService.ts
```diff
@@ -26,33 +28,51 @@ export class AssetsService implements IAssetsService {
...
-    const transaction = await this.AssetsRepository.getNativeTransferCall(param);
-    const hash = await this.wallet.signAndSend({
-      extrinsic: transaction,
-      senderAddress: param.senderAddress,
-      successMessage: param.successMessage,
-    });
-    param.finalizedCallback(String(hash));
+    const useableBalance = await this.AssetsRepository.getNativeBalance(param.senderAddress);
+    const isBalanceNotEnough =
+      Number(ethers.utils.formatUnits(useableBalance, 18)) -
```
bobo-k2, line 33: In `useChainMetadata` you can find `decimals` value, or do what @impelcrypto suggested to avoid passing number of decimals

src/v2/services/implementations/AssetsService.ts
```diff
@@ -28,35 +29,35 @@ export class AssetsService implements IAssetsService {
...
 
   public async transferNativeAsset(param: ParamAssetTransfer): Promise<void> {
+    const { decimal } = useChainMetadata();
     const useableBalance = await this.AssetsRepository.getNativeBalance(param.senderAddress);
-    const isBalanceNotEnough =
-      Number(ethers.utils.formatUnits(useableBalance, 18)) -
-        Number(ethers.utils.formatUnits(param.amount, 18)) <
+    const isBalanceEnough =
+      Number(ethers.utils.formatUnits(useableBalance, decimal.value)) -
+        Number(ethers.utils.formatUnits(param.amount, decimal.value)) >
```
impelcrypto, line 36: Why don't just use `ethers.utils.formatEther(wei)`? So you don't need to use `useChainMetadata`.

src/v2/services/implementations/AssetsService.ts
impelcrypto: @bobo-k2 
Use `ethers.utils.formatEther(wei)`, so that no need to pass decimals.
https://github.com/AstarNetwork/astar-apps/pull/1116#discussion_r1448929113

bobo-k2: Even better, @gluneau I apologize for confusion

impelcrypto (approved):
LGTM

bobo-k2 (approved):

PR #1195, opened 2024-02-19
DApp Promotion – Nova Wallet

Leemo94:


Kahonnohak:
Hi team, thanks for submitting.

A few things
- The image is not right ratio
- The place is at the very end that not visible
- We ask to bring some news for this promotion area. You have a lot to share with awesome videos. I recommend to use that.
Screenshot 2024-02-29 at 11 07 23

PR #1220, opened 2024-03-07
feat: include transaction fee into calculation of ETH balance

impelcrypto:
**Pull Request Summary**

* feat: include transaction fee into calculation of ETH balance 

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Adds**
* feat: include transaction fee into calculation of ETH balance 

=Before=

Some users seem to think the gas fee is too expensive (which appears to be a bug), but it’s just because users don’t have enough ETH, and that’s why MetaMask fails to calculate the gas fee correctly.

image

=After=

Calculate the gas fee and display error if user doesn't have ETH to pay the fee, so that MetaMask won't pop up with wrong transaction fee.

image

image

### DIFF
--- a/src/components/bridge/ethereum/styles/l1-bridge.scss
+++ b/src/components/bridge/ethereum/styles/l1-bridge.scss
@@ -37,7 +37,6 @@
   padding: 10px 20px;
   border-radius: 6px;
   width: 344px;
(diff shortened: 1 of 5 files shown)


gluneau (approved):
It works.

PR #1269, opened 2024-04-25
Burned rewards calculation

bobo-k2:
**Pull Request Summary**

Fixed existing 'tokens to be burned' calculation to use real dApp tiers allocation.

The code goes from the first `era` in the current period 'till `current_era - 1` and calculates how much reward are allocated for empty tiers slots (those are burned rewards). Those calculation is based on real data.

Second part of the calculations calculates the same from `current_era` until last `era` of the current period under assumption that we will have the same dApp tier allocations as in previous `era`. Since we are looking for total tokens to be burned this is the best we can.
The more we are progressing till the end of the current period the more precise calculation will be.

image

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/src/staking-v3/hooks/useDataCalculations.ts
+++ b/src/staking-v3/hooks/useDataCalculations.ts
@@ -115,7 +115,7 @@ export function useDataCalculations() {
 
     // Determine number of used dApp slots per tier.
     for (let dapp of dappTierRewards.dapps) {
-      if (dapp.tierId !== undefined) {
+      if (dapp.tierId) {
         const dappsInTier = slotsPerTier.get(dapp.tierId) ?? 0;
         slotsPerTier.set(dapp.tierId, dappsInTier + 1);
       }
--- a/src/staking-v3/hooks/useDataCalculations.ts
+++ b/src/staking-v3/hooks/useDataCalculations.ts
@@ -1,14 +1,16 @@
-import { computed } from 'vue';
+import { computed, onMounted, ref } from 'vue';
 import { useDappStaking } from './useDappStaking';
 import { useTokenCirculation } from 'src/hooks/useTokenCirculation';
 import { ethers } from 'ethers';
 import { useStore } from 'src/store';
-import { NumberOfStakersAndLockers } from '../logic';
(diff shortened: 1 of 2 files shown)


Kahonnohak:
Thanks @bobo-k2 I'll check & verify by manually adding tokens tomorrow!

src/staking-v3/hooks/useDataCalculations.ts
```diff
@@ -52,20 +54,85 @@ export function useDataCalculations() {
...
+    const slotsPerTier = new Map<number, number>([
+      [0, 0],
+      [1, 0],
+      [2, 0],
+      [3, 0],
+    ]);
+
+    // Determine number of used dApp slots per tier.
+    for (let dapp of dappTierRewards.dapps) {
+      if (dapp.tierId !== undefined) {
```
gtg7784, line 118:
```ts
if(dapp.tierId)
```

I think it is more better for filtering null too

src/staking-v3/hooks/useDataCalculations.ts
```diff
@@ -52,20 +54,85 @@ export function useDataCalculations() {
...
+    // Calculate non-allocated rewards.
+    for (let [key, value] of slotsPerTier) {
+      const slotsPerTier = tiersConfiguration.value.slotsPerTier[key];
+      result += dappTierRewards.rewards[key] * BigInt(slotsPerTier - value);
+    }
+
+    return result;
+  };
+
+  onMounted(() => {
```
impelcrypto, line 133: perhaps add 'async' and 'await' here?

impelcrypto (approved):

src/staking-v3/hooks/useDataCalculations.ts
bobo-k2: I didn't await here because `calculateTotalTokensToBeBurned` internally updates `ref` value and this will cause UI to re-render the value.

src/staking-v3/hooks/useDataCalculations.ts
bobo-k2: fixed

PR #1279, opened 2024-05-05
fix: enables unlock and withdrawal from Lockdrop accounts

impelcrypto:
**Pull Request Summary**

* fix: enables unlock and withdrawal from Lockdrop accounts

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/src/staking-v3/hooks/useDappStaking.ts
+++ b/src/staking-v3/hooks/useDappStaking.ts
@@ -308,7 +308,6 @@ export function useDappStaking() {
       )();
       await stakingService.claimUnlockedTokens(
         currentAccount.value,
-        isLockdropAccount.value,
         t('stakingV3.withdrawSuccess')
       );
     }
--- a/src/staking-v3/hooks/useDappStaking.ts
+++ b/src/staking-v3/hooks/useDappStaking.ts
@@ -22,13 +22,7 @@ import {
 import { Symbols } from 'src/v2/symbols';
 import { ExtrinsicStatusMessage, IEventAggregator } from 'src/v2/messaging';
 import { useStore } from 'src/store';
-import {
-  useAccount,
-  useChainMetadata,
-  useNetworkInfo,
-  useLedger,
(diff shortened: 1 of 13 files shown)


bobo-k2 (changes requested):
I would like to suggest a bit different approach which doesn't change existing codebase that much and it will be easier to clean up later when we stop supporting lockdrop accounts.

I don't think you need to modify dapp staking repository and service since logic you need is implemented in `DappStakingServiceV2Ledger`.
Just change `useDappStaking` to make call to a proper service based on `isLockdropAccount` 
You will also need to introduce `DappStakingServiceV2Ledger` for EVM.

impelcrypto:
Thanks, @bobo-k2 
You are right. The logic can be simplified as you suggest. I've tested on Shibuya and it works well.
Please review again.

bobo-k2 (approved):
LGTM

src/staking-v3/hooks/useDappStaking.ts
```diff
@@ -22,7 +22,13 @@ import {
 import { Symbols } from 'src/v2/symbols';
 import { ExtrinsicStatusMessage, IEventAggregator } from 'src/v2/messaging';
 import { useStore } from 'src/store';
-import { useAccount, useChainMetadata, useNetworkInfo, useLedger } from 'src/hooks';
+import {
+  useAccount,
+  useChainMetadata,
+  useNetworkInfo,
+  useLedger,
+  ETHEREUM_EXTENSION,
```
gtg7784, line 30: I think exporting constant in hooks is not a good way, I think moving `ETHEREUM_EXTENSION` to another files like `src/constants/~~` but I want to know what you thinking about it

impelcrypto: You are right. Please refer to this commit.

gtg7784 (approved):
LGTM! Thank you!

Issue #1259, opened 2024-04-08
[Tokenomics 2.0] Max vs. Realized Inflation Display


Dinonard:
By doing a few simple data fetches, it's possible to show maximum vs. realized inflation.

### 1. Identify New Cycle Start & Max Inflation

When an event from pallet `Inflation, [NewInflationConfiguration](https://astar.subscan.io/event?module=inflation&event_id=NewInflationConfiguration) is emitted, it means a new cycle has started.

A snapshot of the total issuance is taken, and reward pools for the next cycle are calculated based on inflation parameters.
To replicate this, take the block **BEFORE** this even has happened, and calculate the max possible inflation.
```Rust
first_block_in_the_cycle = <block at which event was detected>;
init_total_issuance = query.balances.totalIssuance();
max_inflation_rate = query.inflation.config.maxInflationRate();
max_inflation = init_total_issuance * max_inflation_rate;
```

<img width="1754" alt="Screenshot 2024-04-08 at 15 12 10" src="https://github.com/AstarNetwork/astar-apps/assets/3002868/fb1f5701-3d74-4065-9a10-628bae17f67a">

### 2. Plot Max Inflation Line

The assumption here is that the max inflation rate will be linearly spread over the entire cycle.
X-axis contains blocks, while Y-axis contains total issuance.

The math is fairly simple: 
```Rust
cycle_length_in_blocks = <use runtime API to read this>;

cycle_progression = (x - first_block_in_the_cycle ) / cycle_length_in_blocks; // How much cycle has progressed, always between 0 and 1.
max_inflation(x) = cycle_progression * max_inflation + init_total_issuance;
```
where `x` is the block number.

### 3. Plot Realized Inflation Line

This one is also fairly simple.

At any block within the cycle, we simply read the total issuance:
```Rust
current_block = ...;
realized_total_issuance = query.balances.totalIssuance();
```

Using this number, and knowing the values from the 1st subchapter above, we can estimate the final inflation:
```Rust
cycle_length_in_blocks = <use runtime API to read this>;
cycle_progression = (x - first_block_in_the_cycle ) / cycle_length_in_blocks; // How much cycle has progressed, always between 0 and 1.

slope = (realized_total_issuance - init_total_issuance) / (x - first_block_in_the_cycle);

realized_inflation(x) = (x - first_block_in_the_cycle) * slope + init_total_issuance;
```
where `x` is the block number.

### 4. Expected Result

[...]

PR #1280, opened 2024-05-08
Estimated realized inflation

bobo-k2:
**Pull Request Summary**

Estimated realized inflation calculation ([ref](https://github.com/AstarNetwork/astar-apps/issues/1259)). The calculation calculates slope from first block of the current cycle till the current block and assumes the same slope until the end of the period. This estimation introduces an error which will decrease after stakers clam their rewards (after first build and earn subperiod) and decrease even more as we are approaching the end of the current cycle.

We have beens struggling how to display inflation with just one meaningful number and seems this is the best way so far. In the next iteration we plan to add inflation chart as defined in the document referenced above.

 
image

**Check list**

- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


src/hooks/useInflation.ts
```diff
@@ -1,22 +1,106 @@
...
+    }
+
+    return inflation;
+  };
+
+  watch(
+    [$api],
+    async () => {
+      estimatedInflation.value = await estimateRealizedInflation();
+      console.log('Realized inflation:', estimatedInflation.value);
```
impelcrypto, line 96: Let's use `console.info` if you want to show this in the console

bobo-k2: Actually, this is not needed. Removed in https://github.com/AstarNetwork/astar-apps/pull/1280/commits/e120bdb75d843a397f4821a7f6ebf1ba47b229e8

src/hooks/useInflation.ts
```diff
@@ -1,22 +1,105 @@
...
+   */
+  const estimateRealizedInflation = async (): Promise<number | undefined> => {
+    let inflation: number | undefined;
+
+    if ($api) {
+      try {
+        // Find the block when last NewInflationConfiguration event was emitted.
+        const subscanRepository = container.get<ISubscanRepository>(Symbols.SubscanRepository);
+        const response = await subscanRepository.getEvents(
+          networkNameSubstrate.value.toLocaleLowerCase(),
```
gtg7784, line 42: I think we need to use toLowerCase for this case

gtg7784 (changes requested):

impelcrypto (approved):

gtg7784 (approved):
Thank you! LGTM

PR #1324, opened 2024-06-17
dApp Staking ranked tier update

### LINKED ISSUE #1305: dAppStakingV3 ranked tier update
We're close to merge [PR](https://github.com/AstarNetwork/Astar/pull/1240)
Required changes are minimal.

New property `rank_rewards` is added to DAppTierRewards
Screenshot 2024-06-05 at 10 30 48 PM
Same as tier rewards but contains rank rewards.

New field `rank` is added to event DAppReward
Screenshot 2024-06-05 at 10 29 28 PM

RankedTier type is still u8 like TierId but it contains 2 values (rank & tier). This is done in order to keep PoV low.
Tier exists on the first byte, use bitwise AND to get the value `const tier_id = value & 0xf`
to get rank exists on the second byte, do a right shift 4 bits to get the value `const rank = value >> 4`
You can implement this even before the release because old tier values are from 0 to 3 and the rank will be zero.

DAppStakingApi is bumped to version 2. Nothing is changed here expect the return value which is still the same (u8) but represents different value, which is the encoded RankedTier value, so that's why we bumped it.
Screenshot 2024-06-05 at 10 44 22 PM

***Reward calculation***
Tier reward is the same `rewards[tier]`
Rank reward calculation is very simple `rank * rank_rewards[tier]`
Final amount `amount = rewards[tier] + rank * rank_rewards[tier]`

bobo-k2:
**Pull Request Summary**

This PR is to supported runtime [PR](https://github.com/AstarNetwork/Astar/pull/1240) where dApps will be ranked inside a tier and higher ranked dApps will get more rewards.

Closes #1305 

**Check list**

- [x] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


gtg7784 (approved):
LGTM, thanks!

impelcrypto (approved):

PR #1351, opened 2024-07-08
Inflation chart

bobo-k2:
**Pull Request Summary**

Token burning on Astar happened today and realized inflation chart went wrong. This happened because actual token issuance after burning is lower than at the beginning of staking period 1.
image

As a quick fix I deduced burned amount from token issuance at the beginning of period 1.

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/src/hooks/useInflation.ts
+++ b/src/hooks/useInflation.ts
@@ -78,7 +78,9 @@ export function useInflation(): UseInflation {
       const balancesRepository = container.get<IBalancesRepository>(Symbols.BalancesRepository);
       const initialTotalIssuance =
         (await balancesRepository.getTotalIssuance(period1StartBlock - 1)) -
-        BigInt('350000000000000000000000000'); // Quick fox for token burning event. TODO make a proper solution
+        (networkNameSubstrate.value.toLowerCase() === 'astar'
+          ? BigInt('350000000000000000000000000')
+          : BigInt(0)); // Quick fox for token burning event. TODO make a proper solution
       const realizedTotalIssuance = await balancesRepository.getTotalIssuance();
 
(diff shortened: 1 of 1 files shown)


src/hooks/useInflation.ts
```diff
@@ -76,7 +76,11 @@ export function useInflation(): UseInflation {
       }
 
       const balancesRepository = container.get<IBalancesRepository>(Symbols.BalancesRepository);
-      const initialTotalIssuance = await balancesRepository.getTotalIssuance(period1StartBlock - 1);
+      const initialTotalIssuance =
+        (await balancesRepository.getTotalIssuance(period1StartBlock - 1)) -
+        (networkNameSubstrate.value.toLowerCase() === 'astar'
+          ? BigInt('350000000000000000000000000')
+          : BigInt(0)); // Quick fox for token burning event. TODO make a proper solution
```
gtg7784, line 83: What about not showing if it's Shiden or Shibuya?

gtg7784 (approved):

src/hooks/useInflation.ts
bobo-k2: In that case we will deduce 0n which is fine because no burning happened on Shiden, Shibuya

> What about not showing if it's Shiden or Shibuya?

In that case we will deduce 0n which is fine because no burning happened on Shiden, Shibuya

bobo-k2: > What about not showing if it's Shiden or Shibuya?

In that case we will deduce 0n which is fine because no burning happened on Shiden, Shibuya

PR #1376, opened 2024-08-20
Tier thresholds derivation support

### LINKED ISSUE #1367: dAppStakingV3 - Tier Thresolds Derivation
This [PR](https://github.com/AstarNetwork/Astar/pull/1306) will be merged soon. For additional context, a [forum post](https://forum.astar.network/t/dapps-staking-v3-proposal/4206/170) explains the motivation.

This feature introduces a storage migration from v7 to v8 for pallet dappStaking.

Here are the small breaking changes introduced on types:

#### `TierThreshold` enum is updated

- *DynamicTvlAmount* & *FixedTvlAmount* variants were deleted,
- *DynamicPercentage* & *FixedPercentage* were introduced instead.

```rs
enum TierThreshold {
    FixedPercentage { required_percentage: Perbill },
    DynamicPercentage {
        percentage: Perbill,
        minimum_required_percentage: Perbill,
    },
}
```

*Note - Example of Perbill to Percent transformation: 35_700_000 = 3.57%*

#### `TierConfig` storage value: `TiersConfiguration` type is refactored
- `number_of_slots` field is removed,
- `tier_thresholds` field type is refactored from a vector of `TierThreshold`s to a vector of `Balance`s (u128).

This simplifies tier threshold amounts parsing as now item 0 of the vector refers to tier 1's threshold, item 1 to tier 2's threshold, and so on. The total number of slots can still be compute summing all items in `slots_per_tier` vector.

bobo-k2:
**Pull Request Summary**

Support for dApp staking pallet changes explained in #1367

Runtime changes are not deployed on any network yet, but the code can be tested with local node after pulling the latest `main` branch. The code should work without errors on local node, Shibuya, Shiden, Astar.

Tier threshold value should be visible on UI
image

Closes #1367 

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/src/staking-v3/logic/interfaces/DappStakingV3.ts
+++ b/src/staking-v3/logic/interfaces/DappStakingV3.ts
@@ -1,5 +1,5 @@
 import { Balance } from '@astar-network/metamask-astar-types';
-import { AccountId32, Permill } from '@polkadot/types/interfaces';
+import { AccountId32, Perbill, Permill } from '@polkadot/types/interfaces';
 import {
   BTreeMap,
   Compact,
@@ -126,10 +126,16 @@ export interface PalletDappStakingV3ContractStakeAmount extends Struct {
   readonly tierLabel: Option<PalletDappStakingV3TierLabel>;
 }
 
-export interface PalletDappStakingV3TiersConfiguration extends Struct {
+export interface PalletDappStakingV3TiersConfigurationLegacy extends Struct {
   readonly slotsPerTier: Vec<u16>;
   readonly rewardPortion: Vec<Permill>;
-  readonly tierThresholds: Vec<PalletDappStakingV3TierThreshold> | Vec<u128>;
+  readonly tierThresholds: Vec<PalletDappStakingV3TierThreshold>;
+}
+
+export interface PalletDappStakingV3TiersConfiguration extends Struct {
+  readonly slotsPerTier: Vec<u16>;
+  readonly rewardPortion: Vec<Perbill>;
+  readonly tierThresholds: Vec<u128>;
 }
 
 export interface PalletDappStakingV3TierThreshold extends Enum {
--- a/src/staking-v3/logic/interfaces/DappStakingV3.ts
+++ b/src/staking-v3/logic/interfaces/DappStakingV3.ts
@@ -127,13 +127,12 @@ export interface PalletDappStakingV3ContractStakeAmount extends Struct {
 }
 
 export interface PalletDappStakingV3TiersConfiguration extends Struct {
-  readonly numberOfSlots: Compact<u16>;
   readonly slotsPerTier: Vec<u16>;
   readonly rewardPortion: Vec<Permill>;
-  readonly tierThresholds: Vec<PalletDappStakingV3TierThreshold>;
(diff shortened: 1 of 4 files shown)


src/staking-v3/logic/models/DappStaking.ts
```diff
@@ -172,13 +172,7 @@ export interface TiersConfiguration {
   readonly numberOfSlots: number;
```
ipapandinas, line 172: Perhaps the `numberOfSlots` field can be removed here too.

bobo-k2: I prefer to leave it because it is used on the portal UI. The variable is now calculated as below
`numberOfSlots: configuration.slotsPerTier.reduce((acc, val) => acc + val.toNumber()`

src/staking-v3/logic/interfaces/DappStakingV3.ts
```diff
@@ -127,13 +127,12 @@ export interface PalletDappStakingV3ContractStakeAmount extends Struct {
 }
 
 export interface PalletDappStakingV3TiersConfiguration extends Struct {
```
ermalkaleci, line 129:
```suggestion
export interface PalletDappStakingV3TiersConfigurationLegacy extends Struct {

export interface PalletDappStakingV3TiersConfiguration extends Struct {
```

@bobo-k2 you need 2 different versions here and try to decode with new one and fallback to legacy one. Otherwise this will fail until we perform the migration

gtg7784 (approved):

src/staking-v3/logic/repositories/DappStakingRepository.ts
```diff
@@ -517,21 +523,25 @@ export class DappStakingRepository implements IDappStakingRepository {
...
-            }
-          : {
-              type: TvlAmountType.FixedTvlAmount,
-              amount: threshold.asFixedTvlAmount.amount.toBigInt(),
-            }
-      ),
+      tierThresholds: configuration.tierThresholds.map((threshold) => {
+        // Support both u128 and PalletDappStakingV3TierThreshold.
+        // If threshold has isUnsigned property it's u128.
+        // memo: remove isUnsigned check when u128 is used for all thresholds. Most likely in Astar period 003.
```
impelcrypto, line 532: This prefix should be `Todo` 🙂

impelcrypto (approved):

src/staking-v3/logic/interfaces/DappStakingV3.ts
bobo-k2: I tested new version on local node and on Astar. Decoding works, but I will check how to improve

ermalkaleci: @bobo-k2 It will decode because you're telling the type but the values will be incorrect.

bobo-k2: You are right. Fixed in this [commit](https://github.com/AstarNetwork/astar-apps/pull/1376/commits/24af84ebdc616ada5066928236eec7a9a016f085)

PR #1387, opened 2024-09-19
Fix/selected dapp index

bobo-k2:
**Pull Request Summary**

There were bugs when searching or selecting dApps to vote on. I also applied some fixes suggested by Biome (prefix type import and enum init)

1) if I search for 1 dapps using the SEARCH button. I find it, I select it. Then I want to select another dApp so I delete the name of the previous dapp to write the new search. At that point, the first dApp I had selected changes. (example: if I had selected VLS as the first dapp, after canceling the search I find myself selected as the first dApps destore.)
image
image

2) if I search a Dapp by [CHOOSE dAPP] section, after that I click to [DEFI]. All defi dapps is showed. Now I select 1 or 2 dApps from that list. Now If I want to add one more dApps using the [SEARCH] option, I write name, I find the dapps but the system loose the correct selections of the previous dapps.

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/src/staking-v3/components/vote/VotingWizard.vue
+++ b/src/staking-v3/components/vote/VotingWizard.vue
@@ -74,10 +74,10 @@
 <script lang="ts">
 import { defineComponent, ref, computed, watch } from 'vue';
 import { useI18n } from 'vue-i18n';
-import { WizardItem } from './types';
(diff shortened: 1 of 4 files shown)


impelcrypto (approved):
LGTM

PR #1384, opened 2024-09-16
Fetch block time from chain

bobo-k2:
**Pull Request Summary**

- fetch block time from a chain
- remove hardcoded block time values in code

**Check list**

- [ ] contains breaking changes
- [ ] adds new feature
- [x] modifies existing feature (bug fix or improvements)
- [ ] relies on other tasks
- [ ] documentation changes
- [ ] tested on mobile devices

PR #1389, opened 2024-09-24
APR calculation fix to use block time for a given block

bobo-k2:
**Pull Request Summary**

I made an oversight in #1384 when I forgot to provide block number when fetching block time in APR calculations

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/src/v2/repositories/implementations/SystemRepository.ts
+++ b/src/v2/repositories/implementations/SystemRepository.ts
@@ -85,7 +85,7 @@ export class SystemRepository implements ISystemRepository {
     const api = await this.api.getApi(blockNumber);
 
     if (api.consts.aura) {
-      const blockTime = <u64>api.consts.aura?.slotDuration;
+      const blockTime = <u64>api.consts.aura.slotDuration;
       return blockTime.toNumber() / 1000;
     }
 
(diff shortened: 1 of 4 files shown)


src/v2/repositories/implementations/SystemRepository.ts
```diff
@@ -80,10 +81,14 @@ export class SystemRepository implements ISystemRepository {
     return blockHash;
   }
 
-  public async getBlockTimeInSeconds(): Promise<number> {
-    const api = await this.api.getApi();
-    const blockTime = <u64>api.consts.aura.slotDuration;
+  public async getBlockTimeInSeconds(blockNumber?: number): Promise<number> {
+    const api = await this.api.getApi(blockNumber);
+
+    if (api.consts.aura) {
```
gtg7784, line 87: is it means that the async backing enabled?

bobo-k2: Actually this means that `aura` pallet is not present on chain, which is the case if you ask for block time for some very old block.

impelcrypto (approved):

PR #1400, opened 2024-10-18
Remove staked contracts check

bobo-k2:
**Pull Request Summary**

Removed number of staked contracts check from `canStake` method since we are calling `cleanupExpiredEntries` before staking, if needed.

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/src/staking-v3/hooks/useDappStaking.ts
+++ b/src/staking-v3/hooks/useDappStaking.ts
@@ -446,15 +446,6 @@ export function useDappStaking() {
       } else if (stake.amount <= 0) {
         return [false, t('stakingV3.dappStaking.ZeroAmount'), ''];
       } else if (
-        constants.value &&
-        (ledger.value?.contractStakeCount ?? 0) >= constants.value.maxNumberOfStakedContracts
-      ) {
-        return [
-          false,
-          t('stakingV3.dappStaking.TooManyStakedContracts'),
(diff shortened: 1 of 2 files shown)


gtg7784 (approved):

impelcrypto (approved):

src/staking-v3/hooks/useDappStaking.ts
```diff
@@ -445,15 +445,6 @@ export function useDappStaking() {
...
         return [false, t('stakingV3.dappStaking.ZeroAmount'), ''];
-      } else if (
-        constants.value &&
-        (ledger.value?.contractStakeCount ?? 0) >= constants.value.maxNumberOfStakedContracts
-      ) {
-        return [
-          false,
-          t('stakingV3.dappStaking.TooManyStakedContracts'),
-          docsUrl.dappStakingStaked16Dapps,
-        ];
```
Dinonard, line 456: Just informative - the check isn't completely wrong, just incomplete.
You should also check that the _dapp_ on which the user is trying to stake doesn't have an existing DB entry for the user.

If it does, the call will work, because existing entry is being modified, which won't increase number of staked contracts.
If a new contract is being staked on, one for which the staker doesn't have an entry in the DB, then the call will fail.

PR #1429, opened 2024-12-19
fix: modified logic for checking the native token's remaining balance  

impelcrypto:
**Pull Request Summary**

* fix: modified logic for checking the native token's remaining balance  

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

### DIFF
--- a/src/v2/services/implementations/AssetsService.ts
+++ b/src/v2/services/implementations/AssetsService.ts
@@ -1,6 +1,5 @@
 import { ethers } from 'ethers';
 import { inject, injectable } from 'inversify';
-import { is } from 'quasar';
 import { endpointKey } from 'src/config/chainEndpoints';
 import { LOCAL_STORAGE } from 'src/config/localStorage';
 import { handleCheckProviderChainId } from 'src/config/web3';
--- a/src/v2/services/implementations/AssetsService.ts
+++ b/src/v2/services/implementations/AssetsService.ts
@@ -30,7 +30,7 @@ export class AssetsService implements IAssetsService {
 
   constructor(
     @inject(Symbols.AssetsRepository)
-    private AssetsRepository: IAssetsRepository,
+    private assetsRepository: IAssetsRepository,
     @inject(Symbols.WalletFactory) walletFactory: () => IWalletService,
     @inject(Symbols.CurrentWallet) private currentWallet: string,
     @inject(Symbols.EventAggregator) readonly eventAggregator: IEventAggregator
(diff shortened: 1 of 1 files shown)


src/v2/services/implementations/AssetsService.ts
```diff
@@ -38,23 +39,27 @@ export class AssetsService implements IAssetsService {
...
   public async transferNativeAsset(param: ParamAssetTransfer): Promise<void> {
-    const useableBalance = await this.AssetsRepository.getNativeBalance(param.senderAddress);
-    const isBalanceEnough =
-      Number(ethers.utils.formatEther(useableBalance)) -
-        Number(ethers.utils.formatEther(param.amount)) >
-      REQUIRED_MINIMUM_BALANCE;
+    const isNativeToken = param.assetId === idAstarNativeToken;
+    // Memo: Check if the native token's remaining balance is enough to pay the transaction fee
+    if (isNativeToken) {
+      const useableBalance = await this.AssetsRepository.getNativeBalance(param.senderAddress);
```
bobo-k2, line 45: `AssetsRepository` should be renamed to `assetsRepository` since it is a private variable

src/v2/services/implementations/AssetsService.ts
```diff
@@ -38,23 +39,27 @@ export class AssetsService implements IAssetsService {
...
+      const isBalanceEnough =
+        Number(ethers.utils.formatEther(useableBalance)) -
```
bobo-k2, line 47: There are unnecessary conversions here. The main reason for that is `useableBalance` and `param.amount` are strings. IMO using string for balances is not natural (more natural is `bigint`) and leads to increased code complexity.
You don't need to fix this now, but we should have this in mind for the future

bobo-k2 (changes requested):

src/v2/services/implementations/AssetsService.ts
impelcrypto: Noted, I've modified like [this](https://github.com/AstarNetwork/astar-apps/pull/1429/commits/a60c78c9ebd58c736b53cd18e81b81967df0d974) but we need to do some refactoring in the future.

src/v2/services/implementations/AssetsService.ts
impelcrypto: Fixed in [this](https://github.com/AstarNetwork/astar-apps/pull/1429/commits/a60c78c9ebd58c736b53cd18e81b81967df0d974) commit

bobo-k2 (approved):

gtg7784 (approved):

PR #1492, opened 2025-07-25
Upgrade Polkadot packages to the latest versions

bobo-k2:
**Pull Request Summary**

Updated all Polkadot packages to latest versions.
Motivation:
- chain errors stopped being propagated to ui logic and because of that the portal displayed all transactions as successful, even if they failed
- we were 4 @polkadot/api major versions behind

There is one nasty cast to `any` in Polkasafe code. Any suggestion how to fix is welcome. That's only idea I had since Polkasafe didn't update their package for a very long time and relies on very old @polkadot/api 

Downside of this upgrade is that the project build time is slower than before 👎 

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes


gtg7784 (approved):
LGTM for code changes

bobo-k2:
Thank you, please do some testing when you find a time.

PR #1504, opened 2025-09-29
fix: display `Available amount after staking` correctly

impelcrypto:
**Pull Request Summary**

* fix: display Available amount after staking correctly

**Check list**

- [ ] adds new feature
- [ ] relies on other tasks
- [ ] documentation changes

**Fixes**

The fix now uses `remainingLockedTokensInitial` instead of the reactive `remainingLockedTokens.value`. This ensures that:

  1. The initial locked tokens value is captured before any staking amount is entered (this happens in the watch at the bottom of this file)
  2. When calculating `availableToVoteDisplay`, we use this fixed initial value
  3. This prevents the double-counting issue where the staking amount was reducing both `remainingLockedTokens` and being subtracted again in the display calculation

  Now when you enter a staking amount:
  - The available balance will simply decrease by that amount
  - The remaining locked tokens display will update correctly
 
=Before=

* The `Available amount after staking` should be 558 as the inputted amount is 100, but the result is 458, it decreased the amount twice.

image

image

=After=

image

### DIFF
--- a/src/staking-v3/hooks/useVote.ts
+++ b/src/staking-v3/hooks/useVote.ts
@@ -81,9 +81,17 @@ export function useVote(dapps: Ref<DappVote[]>, dappToMoveTokensFromAddress?: st
       return availableToMove.value - totalStakeAmount.value;
     }
 
-    return remainingLockedTokens.value >= BigInt(0)
-      ? BigInt(useableBalance.value) + remainingLockedTokens.value + availableToMove.value
-      : BigInt(useableBalance.value) - abs(remainingLockedTokens.value) + availableToMove.value;
+    // Calculate available balance after staking
+    // Use the initial remaining locked tokens (before staking amount input)
+    // to avoid double-counting the staking amount reduction
+    const availableBalance = BigInt(useableBalance.value);
+    const initialRemainingLocked = max(remainingLockedTokensInitial, BigInt(0));
+    const actualAvailable = availableBalance + availableToMove.value + initialRemainingLocked;
+
+    // Return the balance after subtracting the staking amount
+    const afterStaking = actualAvailable - totalStakeAmount.value;
+
+    return afterStaking;
   });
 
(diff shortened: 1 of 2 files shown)


src/staking-v3/hooks/useVote.ts
```diff
@@ -81,9 +81,17 @@ export function useVote(dapps: Ref<DappVote[]>, dappToMoveTokensFromAddress?: st
...
     }
 
-    return remainingLockedTokens.value >= BigInt(0)
-      ? BigInt(useableBalance.value) + remainingLockedTokens.value + availableToMove.value
-      : BigInt(useableBalance.value) - abs(remainingLockedTokens.value) + availableToMove.value;
+    // Calculate available balance after staking
+    // Use the initial remaining locked tokens (before staking amount input)
+    // to avoid double-counting the staking amount reduction
+    const availableBalance = BigInt(useableBalance.value);
+    const initialRemainingLocked = max(remainingLockedTokensInitial, BigInt(0));
```
gtg7784, line 88: isn't it max(bn(0), bn(0)) ?

gtg7784: I mean `initialRemainingLocked` is required?

impelcrypto: `remainingLockedTokensInitial` is remaining locked balance, for example this case is 540.699ASTR. It's updated [here](https://github.com/AstarNetwork/astar-apps/blob/20ea62a0351e3e3f6965116c7616b218dd3680c9/src/staking-v3/hooks/useVote.ts#L239).

image

impelcrypto: This is required as to calculate the actual available balance.

> isn't it max(bn(0), bn(0)) ?

No, the value is not `bn(0)`

image

image

bobo-k2 (approved):

gtg7784 (approved):

